PR #11, opened 2015-04-01
Basic - 1 Translation

brymz:
Adds two r markdown and two .html files that are translations of the exercise, Basic-Python-1. The .html is the `knit` result of the r markdown file. The two r markdown files differ in the way they present the data as a list or matrix. List is more directly translated from Python, while matrix is more R flavor. Could also do data frame. Files are inserted somewhat naively into the exercises folder. 

Adds a markdown/html pair for the R translation of Basic-Python-2. Encountered an issue trying to split up comments and code chunks within a for loop. Pushed the whole for loop together for this commit. I made some comments in this file regarding lesson plan. Two extra points exist in this lesson, functions and data management classes, that I comment on in the file and would like to discuss further.

I also want to chat about the `git config push.default`. My two options are `simple` or `matching`.


ethanwhite:
Thanks for this Zack. This will really help kick off our conversation about how to handle a lot of different things. I look forward to chatting about it tomorrow.

brymz:
This pull request has been significantly modified. All original committed files have been deleted. The replacements should reflect our new organizational structure. 
In exercises, I added Basic-1-R.md. The only difference from the Python version is the data that takes on the R flavor of two vectors. 
I added a solutions folder and included Basic-1-R.txt that has the code solution pasted from the console. I have a folder on my local machine that includes the .R code solution. I will push to separate repository when it's created.

exercises/Basic-1-R.md
```diff
@@ -0,0 +1,27 @@
...
+layout: exercise
+title: Basic 1
+language: R
+---
+
+The number of birds banded at a series of sampling sites has been counted by
+your field crew and entered into the following list. The first item in each
+sublist is an alphanumeric code for the site and the second value is the number
+of birds banded. Cut and paste the list into your assignment and then answer the
+following questions by printing them to the screen.
```
ethanwhite: These aren't lists now, they are vectors, and there are two of them. The language should be updated to reflect this.

solutions/Basic-1-R.txt
```diff
@@ -0,0 +1,11 @@
...
+
+[1] "30  birds were counted at the 7th site."
+
+[1] "32  birds were counted at the last site."
+
+[1] "A total of  955  birds were counted across all sites."
+
+[1] "The average number of birds is  37  per site."
+
+[1] "A total of  248  birds were counted at sites with codes beginning with C."
```
ethanwhite: Let's skip the sentences here and just print out the numbers representing the actual results.

exercises/Basic-2-R.md
```diff
@@ -0,0 +1,42 @@
...
+    for the individual, the second column contains the string 'large' or
+    the string 'small' depending on the size of the individuals ears,
+    and the third column contains the GC content of the DNA sequence.
+5.  Prints the average GC-content for both large-eared elves and
+    small-eared elves to the screen.
+6.  Exports the table of individual level GC values to a CSV (comma
+    delimited text) file titled `grangers_analysis.csv`.
+
+
+Built-in functions for R (i.e., read.csv(), matrix(), data.frame()) should be sufficient for this exercise. Throughout the assignment feel free to use whatever data structures you
```
ethanwhite: The examples that I gave in the Python version of the problem have built-in equivalents as you note, but a problem of this size should almost certainly still be broken up into functions (e.g., a function for calculating GC-content), so we should put this part of the language about using functions:
"This code should use functions to break the code up into manageable pieces. "

in in place of "Built-in functions for R (i.e., read.csv(), matrix(), data.frame()) should be sufficient for this exercise."

exercises/Basic-1-R.md
brymz: done.

solutions/Basic-1-R.txt
brymz: done.

exercises/Basic-2-R.md
brymz: done.

ethanwhite:
Looks great! Thanks!

PR #13, opened 2015-04-09
solution Basic-2; both Expressions-variables-1; 2; 3; both Functions-1; exercise Module-1; exercise Expressions-variables-4

brymz:
This adds a bunch of text and markdown files.

Re: Basic-2-R.txt
Do you want the average GC-content for large and small eared elves to be printed on the same line? or identified with text?


exercises/Expressions-variables-4-R.md
```diff
@@ -0,0 +1,28 @@
...
+meter per day, and the total area of the sites, and prints them out.
+
+    site1_g_carbon_m2_day = 5
+    site2_g_carbon_m2_day = 2.3
+    site1_area_m2 = 200
+    site2_area_m2 = 450
+    site1_npp_day = site1_g_carbon_m2_day * site1_area_m2 
+    site2_npp_day = site2_g_carbon_m2_day * site2_area_m2
+    print(site1_npp_day)
+    print(site2_npp_day)
```
ethanwhite: This code chunk should be in back ticks rather than indented.

exercises/Functions-1-R.md
```diff
@@ -0,0 +1,23 @@
+---
+layout: exercise
+title: Functions 1
+language: R
+---
+
+Use the built-in functions `abs()`, `round()`, `sqrt()`, `as.integer()`, `as.numeric()`, and `as.character()` to print out the answers to the following. A built-in function is one
+that you don't need to import a module to use. Use another function, `help()`,
```
ethanwhite: "import a module" -> "load a package" or something else more R'y

exercises/Modules-1-R.md
```diff
@@ -0,0 +1,17 @@
+—-—
+layout: exercise
+title: Modules 1
+language: R
+---
+
+Use the `sqrt()` and `log()` functions from the `math` module, along with the
```
ethanwhite: These are built-ins in R.

ethanwhite:
Looks good other than the few in-line comments I just made. Once you've fixed those let me know and we'll get this in.

exercises/Expressions-variables-4-R.md
brymz: Done.

exercises/Functions-1-R.md
brymz: Done.

'Load a package' is directly from the R documentation. Chose to include
'install' also to be clear that is the first step.

exercises/Modules-1-R.md
brymz: Done.

I chose not to make any changes to this file originally because using
built-in functions for the problem negates its utility as a 'Module'
exercise. We'll have to redesign the problem to make it use a package.

ethanwhite:
Looks great. Thanks!

PR #16, opened 2015-04-17
Strings 1 exercise and solution

brymz:


exercises/Strings-1-R.md
```diff
@@ -0,0 +1,17 @@
...
+title: Strings 1
+language: R
+---
+
+Write a series of print statements that returns the following (include a
+blank line between each answer; ?cat):
+
+1. Post hoc ergo propter hoc
+2. What's up with scientists using all of this snooty latin?
+3. `atgcatgcatgcatgcatgcatgcatgcatgcatgcatgcatgcatgcatgcatgcatgc`. Do this by collapsing a vector (?paste) of 'atgc's established in the following code:
```
ethanwhite: Let's remove this problem from both the problem set and the solution since it's less intuitive to to this in R.

brymz:
Instead of removing the problem, I revised it to serve as an introduction to the three R 'print' functions (print, paste, cat).

PR #17, opened 2015-04-17
Strings 2

brymz:
I've resorted the numbering of the problem to try and build complexity with the function solutions. These problems are not exactly straight forward in R, so we might consider adding some tips. I've included a few already with the link to a functions summary and a call to an R help page.


ethanwhite:
Let's reference `stringr` in this problem since I think it's more straightforward.

brymz:
I envision the built-in exercise as Strings-2 with the stringr exercise as Strings-3. Strings-3 in this case would follow Module 1, as the set up step to install, run, and familiarize with `stringr`. The current Strings-3 would replace Strings-4 (methods) that we won't use.

ethanwhite:
I don't totally follow what you're suggesting, sorry. Are you suggesting two different versions of these same exercises, one with stringr and one without?

brymz:
I've already split the exercises. They are not exact duplicates, but focus
on the couple of things that each (built-in, stringr) do well. They are
both named String-2-R right now with an `_` tag to mark the topic. Happy to
chat about it tomorrow.

ethanwhite:
I like the general plan of `Modules 1` leading into `stringr`. I don't personally think it's worth teaching the built-ins since there's a simpler more straight forward way. The other argument for skipping the built-ins is that Data Carpentry has decided to focus on teaching Hadleyverse modules vs. base R in cases where there are solutions to both. The idea is that we should focus on teaching novices the things that give them the simplest, biggest, payoff most quickly.

PR #20, opened 2015-04-17
add Strings-4-R.md PASS: no methods for string manipulations

brymz:
Methods are sparse in R. The set I am most familiar with is `summary.lm()` for working with linear model components. Couldn't find any methods in basic functions, like character manipulation as in the Python context.


ethanwhite:
Yeah, so let's just not include things like this if they don't apply and show up commonly in introductory level R programming.

PR #26, opened 2015-04-23
Strings 6

brymz:
We decided not to use sentences as output, so I ask to store results in a matrix. The .txt solution is the result if you call the matrix to the console. Do we need to ask students to do that directly in their code?


ethanwhite:
It doesn't look like the original problem asks for the sequence number to be printed out, so to keep this simply I'd suggest just skipping the matrix and just print out the GC contents in order.

PR #36, opened 2015-05-04
Functions 6

brymz:
I was able to track down the difference in `print()` output as noted in the 'NB' (Functions-6-R.md, Line 22).
Functions-5-R.txt has multiple lines with [1]. Functions-6-R.txt is only one line with output separated by a space.

I don't know why. Any thoughts before I go searching?


exercises/Functions-6-R.md
```diff
@@ -0,0 +1,22 @@
+---
+layout: exercise
+title: Functions 6
+language: R
+---
+
+This is a follow up to [Strings 6]({{ site.baseurl }}/exercises/Strings-6/).
```
ethanwhite: Does this need to be `Strings-6-R`

exercises/Functions-6-R.md
```diff
@@ -0,0 +1,22 @@
...
+
+A colleague has produced a file with one DNA sequence on each line. Download
+[the file]({{ site.baseurl }}/data/dna_sequences_1.txt) and load it into R using
+`read.csv()`. The file has no header and is separated by white space (`’’`).
```
ethanwhite: I'm not sure what this means. Each sequence is one a single line. What is the "separated by whitespace"?

exercises/Functions-6-R.md
```diff
@@ -0,0 +1,22 @@
...
+
+Calculate the GC content\* of each sequence by writing a function that takes a
+dna sequence as input and returns the GC-content of that sequence. Print the
+result for each sequence. Before we knew about functions we had to take each dna
+sequence one at a time and then rewrite or copy-paste the same code to analyze
+each one. Isn't this better?
+
+\* The GC content is the percentage of bases that are either G or C (as a
+percentage of total base pairs).
```
ethanwhite: Instead of the * what about just add the description immediately following the original mention:

"Calculate the GC content, the percentage of bases that are either G or C as a percentage of total base pairs,  of each sequence. Do this by writing a function that takes a dna sequence as input and returns the GC-content of that sequence."

ethanwhite:
> Any thoughts before I go searching?

Nope, that's really weird, especially since

```R
a = c(1.1, 2.2, 3.3)
for (val in a){print(val)}
```

prints them one per line as you'd expect.

exercises/Functions-6-R.md
ethanwhite: Why specify a separator at all? `read.csv("dna_sequences_1.txt", header=FALSE)` seems to work fine for me.

brymz: >> `read.csv("dna_sequences_1.txt", header=FALSE)` seems to work

For me too. I added the sep = '' as a fix somewhere, but it's clearly not
needed here.
Obviously, `sep = ','` is used a ton.

ethanwhite: Yeah, I think there is a space delimited file somewhere that needs this. In this case since there aren't separators on each line you could actually use anything for `sep`, so I think it's clearer and lets us get rid of the extra language to just not use it.

brymz: >> lets us get rid of the extra sep language
Done.

>> Instead of the * what about just add the description immediately
following the original mention:
Removed * and rewrote paragraph for clarity.

>> Does this need to be Strings-6-R
Done.

Issue #42, opened 2015-05-05
Translate Loops-4 & Loops-5


brymz:
(Subissue #1)

Loops-4 seems like a useful extension of Loops-2 (old name: 'Loops-3')
Not sure what Loops-5 is about.

ethanwhite:
Loops-5 is a stub for something I wanted to add eventually. No need to translate.

Loops-4 is designed to teach the standard:

```python
results = []
for stuff in things:
    results.append(dostuff(stuff))
```

concept, for which I'm sure there's an analog in R.

brymz:
This abstract code chunk is brilliant! I've got to use it in my workshop this summer.

PR #44, opened 2015-05-06
Loops 3

brymz:
To close issue #42.

I think the way this and the previous Loops problems are written is more appropriate for data.frames and/or `dplyr` exercises. I don't have a good idea yet how to get loops involved.

### DIFF
--- a/exercises/Loops-3-R.md
+++ b/exercises/Loops-3-R.md
@@ -1,0 +1,20 @@
+---
+layout: exercise
+title: Loops 3
+language: R
+---
+
+This is a follow up to [Loops 2]({{ site.baseurl }}/exercises/Loops-2-R)
+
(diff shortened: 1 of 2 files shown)


ethanwhite:
> I think the way this and the previous Loops problems are written is more appropriate for data.frames and/or dplyr exercises. I don't have a good idea yet how to get loops involved.

It's true that you wouldn't normally solve this in a loop. You also wouldn't normally do this in Python. Part of my previous thinking has been to teach the fundamentals and then show folks easier ways to accomplish these tasks using higher level libraries. However, Data Carpentry is more focused on putting the most useful high level libraries in folks hands than on the fundamentals, so I agree that we may not want to end up using some of the looping exercises. If you see things that you think we won't want to use feel free to leave them out of the first round of translating.

PR #50, opened 2015-05-12
Scientific 1

brymz:
rename 'Scientific-Python-1.md' to 'Scientific-1-R.md'


exercises/Scientific-1-R.md
```diff
@@ -0,0 +1,53 @@
...
+Fortunately this dataset has data on the mass of recently extinct mammals as
+well as extant mammals (i.e., those that are still alive today). Take a look at
+the [metadata](http://www.esapubs.org/archive/ecol/E084/094/metadata.htm) to
+understand the structure of the data. One key thing to remember is that species
+can occur on more than one continent, and if they do then they will occur more
+than once in this dataset. Also let's ignore species that went extinct in the
+very recent past (designated by the word 'historical' in the 'status' column).
+
+Import the data into R. If you've looked at a lot of data you'll realize
+that this dataset is tab delimited. Use the argument `sep = ‘’` in `read.csv()` to properly format the data. There is no header row.
```
ethanwhite: `sep = ‘’` won't actually work because those are special quote characters (at least it fails for me). Regardless I think it's better to be explicit here and use `sep = '\t'`.

brymz: >> use sep = "\t".
Done.

Also, updated language of items 1 and 5 to match solution.

PR #52, opened 2015-05-18
Scientific 2

brymz:


exercises/Scientific-2-R.md
```diff
@@ -0,0 +1,36 @@
...
+the file `continent_mass_differences.csv`. If you notice anything
+strange think about what's going on and present the final data in the way that
+makes the most sense to you.
+
+Because you’re working with real data, some things might look strange that are
+not results but barriers to analysis. These inconsistencies in the data will 
+require some cleaning. 
+
+The unknown value used in the dataset is `-999`. R assumes your unknown value is 
+`NA`, but `“NA”` in the data is the code for North America.
```
ethanwhite: Change the fancy quotes to regular quotes in case students copy and paste this.

ethanwhite: This should be done throughout this and the next paragraph.

exercises/Scientific-2-R.md
```diff
@@ -0,0 +1,36 @@
...
+strange think about what's going on and present the final data in the way that
+makes the most sense to you.
+
+Because you’re working with real data, some things might look strange that are
+not results but barriers to analysis. These inconsistencies in the data will 
+require some cleaning. 
+
+The unknown value used in the dataset is `-999`. R assumes your unknown value is 
+`NA`, but `“NA”` in the data is the code for North America. 
+Use the additional arguments `stringsAsFactors = FALSE, na.strings = “”` in
```
ethanwhite: Is there any reason not to use `na.strings="-999"` here to also take care of the -999?

exercises/Scientific-2-R.md
brymz: >> Change the fancy quotes to regular quotes
So, you want me to remove the back quotes everywhere? What about around the
function `str_replace_all()` and arguments `stringsAsFactors = FALSE,
na.strings = “”` in  `read.csv()`? I wrote this paragraph in response to
our discussion of "" and `` last week.

exercises/Scientific-2-R.md
brymz: >> Is there any reason not to use na.strings="-999" here to also take care
of the -999?

When I used na.strings="-999", R inserted NAs in the places with -999, but
I could not remove those NAs using the standard argument `na.rm = TRUE`. I
could certainly look into it more if we feel it will make the solution much
more intuitive.

exercises/Scientific-2-R.md
ethanwhite: > you want me to remove the back quotes everywhere

No, not the back quotes. You currently have `“NA”`, but the double quotes in here are not regular double quotes, and they will error if you enter them into R:

```
> “NA”
Error: unexpected input in "�"
```

They need to be replaced with regular double quotes: `"NA"`

brymz: >> They need to be replaced with regular double quotes: "NA"

Glad I clarified. My text editor calls them 'smart quotes'. I disagree.
'Smart quotes' removed.

brymz: >> 'Smart quotes' removed
AND language changed in unknowns paragraph to use `na.strings = "-999"`.

brymz: >> I wonder if this isn't better solved by using more of dplyr and tidyr?

Updated the comments to include `dplyr` language and the resulting solution.

PR #60, opened 2015-05-27
Scientific 0

brymz:
Significant departures from the Python version:
- change the emphasis to data familiarization
- remove problems to calculate volume and biomass (duplicated in 'Combining Basics')
- Introduce data management with `dplyr`


exercises/Scientific-0-R.md
```diff
@@ -0,0 +1,43 @@
...
+conduct the analysis for her (she might be a world renowned expert in carbon
+storage in plants, but she sure doesn't know much about computers). She has
+placed a data file on the web for you to
+[download]({{ site.baseurl }}/data/shrub_volume_experiment.csv).
+
+You want to get familiar with the data so…
+
+1. Import the shrub dimensions data using `read.csv()`. It has a header row so 
+you'll need to tell R there are column names by providing the optional argument 
+`head = TRUE`.
```
ethanwhite: Seems like read.csv() assumes there is a header, so we shouldn't need the second sentence.

brymz: >> Seems like read.csv() assumes there is a header, so we shouldn't need
the second sentence.
Sentence removed.

PR #77, opened 2015-06-04
Debugging

brymz:
Some significant departures from Python code. Might reconsider what 'bugs' we want to have in buggy-granger-analysis-code-R.R


ethanwhite:
I'll go ahead and merge this version, but feel free to introduce whatever kinds of bugs work best. The only goal here is to get folks using the debugger while still having some bugs for the test after the code starts executing.

PR #91, opened 2015-06-22
Assignments

brymz:
exercises-Python.txt, exercises-R.txt, and exercises-Sql.txt contain tables for the assignment lists. Missing exercises for Python are noted by a 0 in the assignments column. I'll be using these lists to organize the assignment.md, so I figured I'd put them up as reference.

See issue #92.


brymz:
PR moved to #93.

PR #93, opened 2015-06-23
Assignments With _includes

brymz:
(Related PR #91)

Some how, my local 'assignments' branch has gone from 'Assignments' to 'assignments'. They match aside from the newest commit (613a00c). I've closed #91.


ethanwhite:
Looks really good in general.

All of the `redirect_from` lines should be removed. These lines were redirecting paths that used to exist on the old Drupal based version of programmingforbiologists.org that didn't exist in the Jekyll based version of the site. Since this is a new site we don't need redirects of any kind.

Also, instead of listing out the exercise manually, how about we modify `exercises/index.md` so that it separately shows Python, R, and database exercises. This should be doable by adding appropriate headings, each with a modified copy of the existing code that only keeps exercises from the appropriate language.

brymz:
>> All of the redirect_from lines should be removed. 
Done.

>> how about we modify exercises/index.md so that it separately shows Python, R, and database exercises.

Done. I also added anchors to help organize the page. I tried organizing the links in a table, but was having trouble getting the Liquid code to pass a link. Instead, the table was populated by an unordered list of `<li><a href= "{{...</a></li>`

ethanwhite:
Looks great!

We should be able to remove the `exercises-*.txt` files now right? Once that's taken care of this is ready to go in.

brymz:
> remove the exercises-*.txt files

Done.

ethanwhite:
Thanks @brymz! This is really coming along nicely.

PR #96, opened 2015-06-25
About

brymz:


about/why-r.md
```diff
@@ -0,0 +1,31 @@
...
+    research. This also allows students in the class to program on any
+    computer they want without dealing with licensing hassles, and
+    guarantees that anyone can take advantage of the online resources
+    that are being developed.
+4.  It runs on all major operating systems.
+5.  It is used by many scientific programmers and taught in many research
+    science departments. This makes collaborating with experts
+    on large or complex projects much easier.
+6.  [Data Carpentry](http://datacarpentry.org), a truly
+    excellent resource for learning scientific programming, is available
```
ethanwhite: Since this is going to be a Data Carpentry product we should probably pull "a truly excellent resource..." as it might come off a bit oddly.

brymz: > pull "a truly excellent resource..."

Done. I also removed the Jekyll's prefab about.md.

PR #99, opened 2015-06-25
sidebar.html: fix home link

brymz:
Fixed the `Programming for Biologists` link by adding a `/` similar to `Home`.
_config.html::site.base.url  = /datacarp-semester-biology

I think this breaks the `sidebar-nav` class. It's minor, but the `Home` link does not bold when you are on the homepage as it does for the others. This is a minor glitch for how many `/` we'd have to delete throughout the rest of the site. Also, it seems like if you don't mind not having that feature, we can have a lot more flexibility with how we structure the sidebar.

### DIFF
--- a/_includes/sidebar.html
+++ b/_includes/sidebar.html
@@ -2,7 +2,7 @@
   <div class="container sidebar">
     <div class="sidebar-about">
       <h1>
-        <a href="{{ site.baseurl }}">
+        <a href="{{ site.baseurl }}/">
           {{ site.title }}
         </a>
       </h1>


ethanwhite:
What wasn't working about the previous approach?

brymz:
Clicking on 'programming for biologists' at the top of the sidebar didn't
go anywhere just using the link to `{{ site.base.url }}`. Otherwise,
nothing was broken by requiring the preceding `/` except that the sidebar
nav functions are inconspicuously buggy for the 'home' page.

ethanwhite:
Thanks. Looks like the lack of bolding of the home page wasn't working anyway (based on my testing of the current master), so let's go ahead and get this in. If you think it's worth taking out the highlighting to do other things with the sidebar definitely go ahead and submit a PR for the change that makes good use of dropping it and I'll take a look.

Issue #103, opened 2015-06-28
Change site title to Data Carpentry for Biologists


ethanwhite:

brymz:
Done. #105

PR #105, opened 2015-06-29
_config.yml: update title

brymz:
(As per #103)

I chose to add '-' around the 'for' for visual appeal. I don't feel strongly about it though. There might be benefits to keeping it simply 'Data Carpentry for Biologists'. The two options I considered are:

Data
Carpentry
-for- Biologists

OR

Data
Carpentry for
Biologists


ethanwhite:
We can always change it back later :)

Issue #110, opened 2015-06-30
Issues with formatting in assignments


ethanwhite:
Something got mixed up a little in the formatting of assignments. Compare:

* http://www.programmingforbiologists.org/assignments/programming-2/
* http://ethanwhite.github.io/datacarp-semester-biology/assignments/Python-2/

I think this is happening because the assignment name is capitalized. See the commit message for my solution to this:
https://github.com/ethanwhite/progbio/commit/79c5bbc44b8da4b988c9ad6bd41ffb0eda51a546

This means that the assignments need to start with lower case letters and the exercises start with capital letters. Yes, it is awful.

brymz:
>> assignments need to start with lower case letters

After I lower case the assignments .md files, git does not seem to
recognize the change. There is "nothing to commit". When I `git rm
assignments/Python-1.md` it removes the 'assignments/python-1.md'

brymz:
#113

ethanwhite:
This works fine for me:

```bash
ethan@croryx:~/datacarp-semester-biology (gh-pages)$ git mv assignments/Python-1.md assignments/python-1.md
ethan@croryx:~/datacarp-semester-biology (gh-pages +)$ git status
On branch gh-pages
Your branch is up-to-date with 'origin/gh-pages'.
Changes to be committed:
  (use "git reset HEAD <file>..." to unstage)

	renamed:    assignments/Python-1.md -> assignments/python-1.md
````

ethanwhite:
Are you on a Windows machine by chance? OS X is case sensitive (as is Linux).

brymz:
I'm on OS X. The command line `git mv` worked. My first approach, just
renaming in the finder, was not recognized by git.

brymz:
#114

PR #114, opened 2015-07-06
lower case assignments AND add output

brymz:


brymz:
Pushed to the wrong branch for the Output work I did (#106). I could separate, but figured the fix to #110 was straight forward enough to just leave it.

Issue #75, opened 2015-06-01
Jeckyl Formatting


brymz:
(For down the road.)

I saw the newest Software Carpentry lesson (http://swcarpentry.github.io/web-data-python/) and it made me think about the way we format our exercises and how that will look when rendered by Jeckyl.

I'd like to look through a couple examples to gather some thoughts and chat with you sometime. We can also look through some of the Data Carpentry lessons, though they like look they are still mostly 'generic' github wiki pages.

brymz:
One version of syllabus and schedule.
Assignments - ?
index.md in root - update to one 'course'
exercises/index.md - auto generate
_includes/sidebar - will change to match new structure.
_layouts - change formatting for different page types

brymz:
Some thoughts after looking through the Jekyll documentation...I really like how they flag important things. However, I felt I didn't get much out of the content in between because they were so commanding. This is similar to my thoughts on how many different types of flags / box types there are in Software Carpentry lessons. My take away is that it is important to use flags like road markers that tell you where you are or what is interesting without you driving off the road.

brymz:
"2nd Pass" Assignment Formatting is described by issues with the "formatting" label.
Issues: #31, #35, #53, #55, #62, #79, #95.
R PRs : #116, #117, #124, #125, #129, #130,

brymz:
I think we've sufficiently addressed this. If we decide in the future to overhaul the formatting of the website we can open a new issue.

PR #131, opened 2015-07-09
2nd Pass R Assignment 6

brymz:
(As per #75) 
Major renaming and file shuffling to get titles in numerical order.


ethanwhite:
Did you use `git mv` for this renaming or did you do it some other way? I ask because the diff here is really confusing since it looks like massive changes are being made to the exercises when they're really just being renamed. This might be unavoidable since you're renaming them to the names of other existing files, but if it is avoidable we should try to do so.

brymz:
I did not use `git mv`, mostly because I needed to move a set of files to a
place holder so they didn't 'already exist'. Should I try to make the file
move and edits separate PRs? Might run into merge problems that way.

ethanwhite:
Ah, right, that makes sense. Can you explain a little more about why you needed to do the renames. We really don't want the build to be dependent on having the order in the assignment match the numbering anyway.

brymz:
The original names were 'Graphing 2', 'Graphing 3', 'Graphing 1'. I think
this would seem really odd to someone who was trying to work with our
material. Most of the titles are already in numeric order now because I
cleaned them up as I was going along (e.g., Expressions and Variables,
Making Choices). Expressions and Variables and Making Choices dropped
unused exercises and condensed title order, which gave me the place holder
position when renaming. I chose to make the rename for Graphing in the
second pass to push off the ugliness from all of the direct overlap in
names with no obvious place holder to a point when the site was getting
established.  The build is still completely dependent on exercise title,
which happens to be in numeric order when appropriate for ease of use.

ethanwhite:
OK, got it. Let's just eat the ugly history on this one and get it done. Thanks!

brymz:
Thanks. :: sigh :: Tried to keep it all in one place.

PR #138, opened 2015-07-10
2nd Pass Sql Assignment 1

brymz:


ethanwhite:
Seems to be rendering fine on GitHub.

ethanwhite:
Comment from Zack (moved out of PR body):

> Using both TextEditor and Atom, I had a difficult time getting the numbered list in Sql-import-data.md to render cleanly. In Atom, some of the numbers were black, while others showed as brown.

ethanwhite:
And the rendering looks good on the site:
http://ethanwhite.github.io/datacarp-semester-biology/exercises/Sql-importing-data/

so we're all set.

brymz:
Yep. Got it to work fine, but the .md file has 'extra' line breaks  in places to get that result.

ethanwhite:
Ah, yes, it's not uncommon for that to be necessary since markdown parsing, in general, will merge lines without an extra line break between them. Sometimes the parses can handle it properly from the context, but not always.

PR #155, opened 2015-08-17
Syllabus and schedule cleanup

ethanwhite:
1. Gets rid of extra syllabi, schedules, and landing pages left over from when class was split into 3
2. Moves the remaining syllabus and schedule into the root directory since the simplicity created by (1) removes the need to put these core files in a separate folder.
3. Gets rid of old redirects that were only needed when this was programmingforbiologists.org.

@brymz - time to turn the tables and have you reviewing at least a few PRs from me :smile:


ethanwhite:
Just updated with a couple of additional cleanup commits.

brymz:
Is there a standard PR review procedure?

ethanwhite:
Probably not.

My process is:

1. Read PR message and maybe glance at `Commits` tab to get an overview
2. Click on `Files changed` tab, read changes
3. If the changes are complicated and it's not obvious that they'll work I might checkout the PR and test them. This tends not to be as generally relevant in cases like this where it's mostly changes to the markdown, but can be important for code.
4. If everything looks good - merge, otherwise provide feedback via either general or in-line comments, which ever I think is most helpful given the issue.

brymz:
For future reference, how do you 'checkout the PR'?

ethanwhite:
This is how I do it:
https://gist.github.com/piscisaureus/3342247

There are other approaches as well.

brymz:
Thanks!

brymz:
Looks good! Go ahead and merge. Have I earned 'write access'?

ethanwhite:
Oops - I'd created a team with you having write access and apparently didn't manage to associate it with this repo. You should be able to press the big green button now, so go ahead and merge and let me know if there are any issues.

PR #158, opened 2015-08-18
Simplifies the R Data Analysis exercise to remove for loop

ethanwhite:
This is intended to be an early exercise, so we can't assume familiarity with
for loops yet. This replaces the for loop with colMeans, which is closer to
the usage of .mean(axis=0) in the Python version.


brymz:
`colMeans()`! That's a new one for me.

PR #160, opened 2015-08-19
Vectors

brymz:
Rename exercises and solutions for Week 5 Assignment (newr-2). Update titles and content.

Vectors-R -> Vectors-1-R
Loops-2-R -> Vectors-2-R
Loops-3-R -> Data-frames-R

There are some extra commits. Data-frames-R.md was briefly Vectors-3-R.md and Data-frame-R.md. Also I missed an edit to the link in a commit to Data-frame-R.md


ethanwhite:
My initial reaction to this is that I'd rather keep the existing exercises and create new exercises as necessary. Some of those exercises might be similar to existing ones (e.g., for loop vs vectorization), but part of the goal here will be to have a pool of exercises that folks can choose from. That said if you see a good reason for this current approach instead just let me know.

brymz:
I changed the approach to solving the Loops-2 and Loops-3 exercises with
vector algebra and data.frames, respectively, in the initial translation.
I'm not sure if vector algebra is a unique feature of R, but the decision
to use vectors instead of a `for` loop was in response to a conversation we
had about choosing language-style solutions instead of direct translations.
The name change from Loops -> Vectors and Data Frames was mostly to catch
the title / file name up to the exercise content.

As for `for` loop exercises, I can easily modify a copy of the problem back
to the `for` loop solution and name that accordingly if you want that
option available. For-loops-2.md might also be a good problem to translate.

ethanwhite:
OK.

Comments:

1. We shouldn't create new assignment names. Once we've fully settled on an order then the exercises for each assignment can be filled in.
2. As I mentioned somewhere else in this flurry I'd prefer something like R Assignment 1, to naming based on weeks.

brymz:
1. I think what you are saying here, is no new assignment file names (like the newr-1.md)? If so, I agree. The newr-1.md is mostly just a place holder so we can have two working drafts of the assignment. It is my intention to return to a final set of r-1.md.
2. In the flurry, I responded to this point in a separate email. And, you just responded to me.

brymz:
Update assignment title.

ethanwhite:
> I think what you are saying here, is no new assignment file names (like the newr-1.md)? If so, I agree. The newr-1.md is mostly just a place holder so we can have two working drafts of the assignment. It is my intention to return to a final set of r-1.md.

Yes, we're on the same page for the goal, just not the process. I don't think we want to create temp files in the history, so we should either just wait to change the assignments until we're ready or go ahead and change them now even if it isn't perfect.

ethanwhite:
Just following up on these two open PRs. I haven't merged them yet due to the temporary assignment names. As soon as those are fixed, I can get this and #161 in.

brymz:
I'm on it.

brymz:
Made the change to remove the temporary assignment file (new-2.md), but looks like I caused a merge conflict. Tried to avoid that by referencing the r-2.md that you added, but I was unsuccessful. Any ideas?

brymz:
Ok. Backtracked for merge. Will update assignment file in a separate PR.

PR #161, opened 2015-08-19
Tables

brymz:


brymz:
Turns out 'Combining Basics' doesn't work at all. It calls for `if`, `for`, AND `function()`.

brymz:
Update assignment title from Week 5 to R Assignment 3.

brymz:
Removed temporary file. This PR now only contains edits to the exercises. Will make a new PR with changes to r-3.md to avoid conflict.

PR #162, opened 2015-08-20
Add draft schedule

ethanwhite:
The old schedule was from Programming for Biologists, this updates it to
include databases and focus more on data in general, as well as moving towards
using more of the existing Data Carpentry curriculum.

### DIFF
--- a/schedule.md
+++ b/schedule.md
@@ -55,13 +55,11 @@ title: Schedule for Programming For Biologists
 * Importing data
 * Basic subsetting
 * Readable code
-  * Functions
 
 #### Reading
 
 * [Starting with data](http://datacarpentry.github.io/R-ecology/02-starting-with-data.html)
 * [Introduction to data frames](http://datacarpentry.github.io/R-ecology/03-data-frames.html)
-* [Creating functions](http://swcarpentry.github.io/r-novice-inflammation/02-func-R.html)
 <!--TODO: Add reading on subsetting data frames -->
 
 [Assignment 4](/assignments/r-2)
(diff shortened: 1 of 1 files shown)


schedule.md
```diff
@@ -3,64 +3,138 @@ layout: page
...
 
 #### Reading:
 
-* [SWC Novice Python: Making Choices](http://software-carpentry.org/v5/novice/python/04-cond.html)
+* [Getting Started](http://datacarpentry.github.io/R-ecology/00-before-we-start.html)
+* [Introduction to R](http://datacarpentry.github.io/R-ecology/01-intro-to-R.html)
 
-Additional Material: 
-[Problem decomposition](http://programmingforbiologists.org/material/problem-decomposition)
+[Assignment 3](/assignments/r-1)
```
brymz: Did we want to change this link to 'R Assignment 1'?

schedule.md
```diff
@@ -3,64 +3,138 @@ layout: page
...
 
-* [Pandas - working with data they way your brain wants](http://nbviewer.ipython.org/urls/github.com/ethanwhite/progbio/raw/master/ipynbs/pandas.ipynb)
-  
-[Assignment 5]({{ site.baseurl }}/assignments/programming-5)
+#### Reading
 
-### Week 6: Graphing, Style, Program Design
+* [Starting with data](http://datacarpentry.github.io/R-ecology/02-starting-with-data.html)
+* [Introduction to data frames](http://datacarpentry.github.io/R-ecology/03-data-frames.html)
+* [Creating functions](http://swcarpentry.github.io/r-novice-inflammation/02-func-R.html)
```
brymz: I wonder about including 'Creating Functions' here. The exercises up through 'Graphing' & 'Statistics' all use built-in functions. So, I thought it would make sense to push off writing functions until after the 'working with data' block. On the other hand, I'm not sure if there is a good resource that separates using functions from writing functions. I think the functions exercises are independent of other material, so they should fit fine with where ever you decide to put them.

ethanwhite:
Yeah, I keep struggling with where to put functions. I like your idea so I've gone ahead and moved them to Week 8. I like assignments numbered with respect to the class not the topic in this case.

PR #163, opened 2015-08-20
Allow customizing Assignments page

ethanwhite:
Individual runs of the course will need to customize the assignments they use. This change
allows the user to set the assignments by providing a list of assignment names in the order
of the assignments. To accomplish this cleanly names of assignments were updated to include the
language of the assignment and an All Assignments page was added so that people considering
using the course or using the material outside of the classroom can see the full list of assignments.

### DIFF
--- a/_includes/assignments.html
+++ b/_includes/assignments.html
@@ -1,0 +1,11 @@
+<ol>
+{% for assignment in page.assignments %}
+  {% for otherpage in site.pages %}
+    {% if assignment == otherpage.title %}
+      <li>
+        <a href="{{ otherpage.url | prepend: site.baseurl }}">{{ otherpage.title }}</a>
+      </li>
+    {% endif %}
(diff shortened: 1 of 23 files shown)


ethanwhite:
Hey @brymz - I know we've gone back and forth a bunch of times about how to juggle the needs of the individual course vs. the needs of folks using the material online. The assignments piece is the hardest to figure out and this is the best I've come up with.

All of my PRs will need to be merged by tomorrow since the start of the semester is upon us. If you have time to review them that's great, but if not I'll just merge them in the morning and keep moving forward. Sorry for this last minute burst, but at least the start of the semester is forcing me to finalize decisions on things :)

brymz:
I think this is a great way to balance ease of use and content! We might want to make it clear somewhere in documentation how _includes/assignment.html is different from _includes/assignments.html.

PR #175, opened 2015-08-24
R Assignment 5

brymz:
... I do call my grandmother every weekend.

### DIFF
--- a/assignments/r-5.md
+++ b/assignments/r-5.md
@@ -2,7 +2,7 @@
 layout: page
 title: R Assignment 5
 language: R
-exercises: ['Scientific 0', 'Scientific 1', 'Scientific 2']
+exercises: ['Functions 2', 'Functions 3', 'Functions 4', 'Functions 5']
 ---
 
 {% include assignment.html %}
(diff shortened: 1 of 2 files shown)


brymz:
This week in the schedule combines functions and loops. I like the idea of adding the problem sequence of 'Strings 6', 'Functions 6', 'Loops 1', condensed as a single problem 'Loops 1'. The probalem asks students to construct a function to calculate GC-content and run it in a loop for a number of sequences. The challenge is that to be able to adequately address the problem, we would also want to introduce the 'stringr' package.

I wanted to get your thoughts before I went ahead and made the changes. I could see us just including the `stringr:str_count()` reference as a hint. Maybe one of the skills we're having students tackle is how/where to find documentation to figure out how to implement a function.

ethanwhite:
I think have a brief introduction to `stringr` as part of a problem is good at this point in the semester, but let's go ahead and give it a few sentences to get them where they need to go. In my experience students end up spending *a lot* of time trying to find documentation on how to implement things during the semester.

ethanwhite:
Should I go ahead and merge this or wait until you've written the new exercise?

brymz:
Merge is ok. Otherwise, leaving it open will help me keep track of my
to-dos when I get back to this at the end of the week.

PR #180, opened 2015-08-24
Put extra R exercises into r-10.md, r-11.md, r-12.md; update all-assignments.md

brymz:


ethanwhite:
This raises an interesting question. I've always thought about the approach here as:

We have a big bag of exercises that folks have written. Some will work for a particular course, some won't, and it's the instructors job to select those and assemble them into assignments. However, there is some value in having pre-assembled assignments the could be chosen from and then (if needed) modified. If we're going that route then it would make more sense to have them named informatively, e.g., `r-assignment-strings.md`.

Thoughts?

brymz:
At one point during this restructuring, I started adding subtitles to the
assignment markdown files. You updated the files and I discarded those
changes in the interest of minimizing extraneous effort. Including
descriptive detail about the exercises is why I added subtitles there. So,
my thoughts...

I like having the simple numeric structure to exercises and assignments for
ease of organization through grouping of content at a higher level. (Of
course, that's broken down somewhat in our most recent restructuring.) I
like the subtitles approach to adding back in the descriptions of content.

ethanwhite:
Subtitles sound good to me. I'll merge this and let you do another round adding the subtitles to everything.

PR #181, opened 2015-08-24
R Assignment 8

brymz:


brymz:
Exercises 'Version Control 7' - 'Version Control 9' are redundant with previous exercises but emphasizing shell commands. Could modify to include interaction with the R translation of `rich_pred.py`.

ethanwhite:
1. I currently have this assignment named `git-1` since it isn't really about `R`. If you see a reason for not changing it to that let me know.
2. I agree that we'll need to tweak some of these later, but think the straight conversion is good for a first pass.

brymz:
Are we going to distinguish between the git assignment for R and the
version written with a Python example? I don't feel strongly about either
name.

ethanwhite:
Good point. I guess let's make it git-r-1 and git-python-1.

brymz:
Done. (Except I chose git-1-R.md to follow the form we use in the
exercises.)

ethanwhite:
Good call. Thanks.

PR #203, opened 2015-09-04
Improve README as a source of information to newcomers

ethanwhite:


brymz:
Do we want to have a README or a clearly marked starting point for
self-guided students?

ethanwhite:
I think that would belong as a link on the front page of the website to a page with some instructions. I agree that something like this is a good idea.

Issue #187, opened 2015-08-27
Add an exercise that loops over multiple files


ethanwhite:
A nice example of looping that applies to a lot of folks' research is to loop a bunch of different files doing the same thing to each file. We should develop a problem around this. If there's something in the Genomics lesson space (see: http://datacarpentry.github.io/lessons/) that works it would be nice to add a genomics exercise or two.

brymz:
Would this be something like the gc-content question in Loops 2, but where
each file is a sequence? In this case, the loop would come before the
function as it calls in the file? Don't know enough about genomics to know
intuitively what you mean by a 'genomics exercise'.

ethanwhite:
> Would this be something like the gc-content question in Loops 2, but where each file is a sequence?

That would work.

>  In this case, the loop would come before the function as it calls in the file? 

I don't quite follow the question.

> Don't know enough about genomics to know intuitively what you mean by a 'genomics exercise'.

Yeah, me either. Mostly what I meant was to take a look at the Genomics lessons on the lessons page and see if there's something easy enough for non-genomics people to understand that we could write an exercise or two around it. In particular I think the exercise you proposed could be made more realistic by using actual sequence file formats instead of just made up strings.

brymz:
Ok. I've got a couple of hours left for the week once I finish adding the
subtitles. Is this what you want me to work on? Or is there something more
pressing? Do we want the subtitles to appear on the site or just be in the
files for reference?

ethanwhite:
On 08/27/2015 01:18 PM, Zack Brym wrote:
> Ok. I've got a couple of hours left for the week once I finish adding the
> subtitles. Is this what you want me to work on? Or is there something more
> pressing?
Sure, that sounds good.
> Do we want the subtitles to appear on the site or just be in the
> files for reference?
I don't have a strong preference a priori.

brymz:
Addressed in #204.

PR #204, opened 2015-09-08
Loops 3

brymz:
(As per #187).


ethanwhite:
This looks great! Nice work. What do you think about boosting the number of data files up to 10 or more to drive home the value of automating this?

brymz:
I added six more files for a total of nine. I decided to stop when the
patricbrc.org data tool started lagging and timing out. Nine seems plenty.

I like the 'value of automating' idea as a concept driven by `for` loops.
There are two places in the lesson that require 'automation' (or `for`
loops); the number of data files, and the size (or number of lines) of each
one. Some are >30,000 lines, so running the analysis takes a minute or two
of processing.

ethanwhite:
Thanks!

Issue #215, opened 2015-09-15
Add exercise that exports data from Portal Database & imports to data frame


ethanwhite:
This is for assignment r-2 and is intended to revisit SQL. This should be added to r-2.

brymz:
For now or next semester?

ethanwhite:
Now (hence the `highpriority` tag).

brymz:
Ah. Tags don't come through in the email. I'll remember that.

brymz:
Do you want the export of the data from the Portal Database to be a SQLite or R operation?

ethanwhite:
SQLite, the class doesn't know how to do it from inside R at this point.

brymz:
Word.

PR #220, opened 2015-09-17
add Data Frames 2; rename first set to Data Frames 1

brymz:
As per #215


ethanwhite:
Looks good in general. The one thing I see is that we want to be clear what is directions and are sub exercises, so let's separate the unordered bulletted list in a such a way as to make clear that the first thing is an instruction and then their are 3 sub-exercises. Also, telling them to use the Console might be confusing. Maybe:

```
Import the 100-Largest-Individuals.csv file and answer the following questions:

1. Use the unique() function to determine how many species are represented in the 100 largest individuals sampled.
2. Use the unique() function to determine how many years are represented in the 100 largest individuals sampled.
2.     Use the min() and max() functions to check how much smaller (as a percent) the 100th largest individual is than the largest.
```

brymz:
> let's separate the unordered bulletted list in a such a way as to make clear that the first thing is an instruction and then their are 3 sub-exercises. 

Done. The update is a near copy of your suggestion. The differences are that I: 1) added back ticks around the file and functions, and 2) modified the statement to 'complete the following tasks:'.

PR #224, opened 2015-09-18
Update Sql-joins-1.md

andrewmarx:
Many people were interpreting 'full species name' to mean genus and species.


ethanwhite:
Actually that's how I intended it and it matches the provided output. That said, I agree that this could be clearer. What do you think about changing the relevant text to:

> full species name (i.e., the genus and species)

to make things more explicit?

andrewmarx:
My bad, I've also been seeing people interpret it as just meaning the species field as well and was thinking of just clarifying it by explicitly listing the fields in the same way that year, month and day are.

ethanwhite:
No problem. It's definitely confusing. I think an update to this PR that made genus and species explicit would be a great addition.

ethanwhite:
Closed by #239.

PR #227, opened 2015-09-22
Assignment R-2 updates

ethanwhite:
Updates and expansions of the exercises in the R-2 assignment.


ethanwhite:
@brymz I'm going to go ahead and merge this, but wanted to make sure you saw it. I've already updated the solutions for Vectors-2, but the solutions need to be updated for Data-frames-1-R and Data-frames-2-R. Let me know if you need any help.

brymz:
For Data-frames-1-R, we've talked previously about avoiding multiple ways
to do things. The built-in function `subset()` is very similar to the
`dplyr` function `filter()` that we introduce later. Is this redundant on
purpose?

ethanwhite:
Yes. I think that students are going to see lots of base R and so probably need to be familiar with it. Either way, it's too late now since class starts in < 30 minutes :smile:

brymz:
Cool. I like your reasoning.

PR #238, opened 2015-09-29
Sql-importing-data.md: update surveys link

brymz:
As per https://github.com/weecology/datacarp-semester-biology-exercise-solns/issues/80#issuecomment-144172240


ethanwhite:
I think you'll also need them to download a new copy of the db as part of your new exercises because otherwise they won't get the new version. This part looks good though, so merging.

PR #242, opened 2015-10-06
Remove statistics problems

ethanwhite:
Drop's statistics problems to focus on data visualization.


ethanwhite:
@brymz - since I'm going to be out of town the second half of this week I decided that the data visualization stuff would be enough. Therefore I removed the two stats questions, so there's less rush in cleaning up the solutions to those. Sorry for all the rushy last minute changes.

PR #243, opened 2015-10-06
Update Graphing 1 & 2 solutions

brymz:


ethanwhite:
In the cases where the axes are logarithmically scaled (i.e., not where the data is log-transformed) then the axes should not be described as Log because the units aren't.

brymz:
Done.

ethanwhite:
Great! Thanks.

Issue #229, opened 2015-09-22
Rename the views in portal_mammals.sqlite to be camel case or underscored


brymz:
As per clarification in #178

PR #244, opened 2015-10-06
rename output view names with _

brymz:
As per #229.

Somehow the `ó` in `Undómiel` was translated in my text editor as a `√>` and is now shown as a `-` in the file changes. I will check back after the merge to see if I need to make a change.


ethanwhite:
This is causing a page build failure. Let's just use an o.

PR #247, opened 2015-10-13
General cleanup of R-5 assignment

ethanwhite:
The major change is shifting Loops-3-R to using Bioconductor instead of manually parsing the fasta files.

### DIFF
--- a/exercises/Functions-2-R.md
+++ b/exercises/Functions-2-R.md
@@ -23,15 +23,16 @@ length in meters for a particular set of parameter values, those for *Theropoda*
 ```
 get_mass_from_length_theropoda <- function(length){
   mass <- 0.73 * length ** 3.63
-  return (mass)
+  return(mass)
 }
 ```
 
-1. Use this function to print out the mass of a Spinosaurus that is 16 m long
+1. Add a comment to this function so that you know what it does.
+2. Use this function to print out the mass of a Spinosaurus that is 16 m long
 based on it's reassembled skeleton. *Spinosaurus is a predator that is bigger,
 and therefore, by definition, cooler, than that stupid Tyrannosaurus that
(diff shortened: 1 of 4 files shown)


ethanwhite:
Merging for class tomorrow.

@brymz - if you can update the answers by class time tomorrow I'd appreciate it as always.

brymz:
From my read of the changed file, it looks like I will only have to update the output solution for Loops-3-R. Is that right?

ethanwhite:
I think that's right.

ethanwhite:
And, now that I think about it let's have them store the output in a data frame (I'm going to teach them how to do this today).

ethanwhite:
I'll change the exercise, you change the solution.

brymz:
Roger that.

brymz:
I need to update my version of R to use `ShortRead`.

ethanwhite:
> I need to update my version of R to use ShortRead.

The class is already all on the most recent version due to something from dplyr (I thought), so thanks for the heads up, but that should be OK.

brymz:
Cool. I'll do my best to wrap up in 15 min. Do you want me to do the merge?

ethanwhite:
Yeah, merge when ready.

brymz:
> install.packages('ShortRead')
Warning in install.packages :
  package ‘ShortRead’ is not available (for R version 3.2.2)

Going to try version 3.2.1, next.

brymz:
> Going to try version 3.2.1

Back on 3.2.2. After all of that it looks like I just used to wrong function call:
It's `biocLite("ShortRead")` not `install.packages("ShortRead")`

ethanwhite:
Yeah, that's right. Glad you got it figured out.

Issue #209, opened 2015-09-08
Link reading/video material to assignments


ethanwhite:
It would be useful to link the reading/video material to the assignments as this would make it easier for self-guided students to find the material relevant to them. If this material was stored in a structured way it could also be used to auto-generate the schedule. `schedule.md` could basically be a YAML list of the assignments to be included and in what order and that would then auto-generate the page.

ethanwhite:
If we go the auto-generated route then the `all_assignments` page could be made to look like:

```
Assignment 1: Reading, Exercises
Assignment 2: Reading, Exercises
...
```

With `Reading` and `Exercises` linked to the the relevant files.

brymz:
This is addressed by #278

PR #263, opened 2015-10-26
WIP: Structure Readings

brymz:
As per #209

### DIFF
--- a/readings/R-2.md
+++ b/readings/R-2.md
@@ -1,20 +1,20 @@
 ---
 layout: page
 title: R Assignment 2
-subtitle: Reading
+subtitle: Readings
 language: R
 ---
 
-__Topics__
+* Topics
 
-* Matrices & data frames
-* Importing data
-* Basic subsetting
-* Readable code
+   * Matrices & data frames
+   * Importing data
+   * Basic subsetting
+   * Readable code
 
-__Reading__
+* Readings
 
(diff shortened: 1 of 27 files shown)


readings/R-2.md
```diff
@@ -0,0 +1,20 @@
+---
+layout: page
+title: R Assignment 2
+subtitle: Reading
+language: R
+---
+
+__Topics__
```
ethanwhite: This should be header.

readings/R-2.md
```diff
@@ -0,0 +1,20 @@
...
+
+* Matrices & data frames
+* Importing data
+* Basic subsetting
+* Readable code
+
+__Reading__
```
ethanwhite: This should be a header.

ethanwhite:
I like the approach. Let me know when it's ready for a full review.

brymz:
I've got through generating the readings files, updating the assignment files, and getting the readings files on the assignments/index.md. I can see assignments/index working as a table, but I don't know how to number rows. We could also add an image/button for `[Assignment]` and `[Reading]` links.

With the integration of readings and assignments, I'm not sure we also need a schedule page. Is there content on the schedule page you want to maintain? 

If you're happy with the restructuring, it should be ready to go. I'm not exactly sure where the merge conflict it (maybe git-1-R assignment title). Happy to take suggestions for the best approach to resolve.

PR #268, opened 2015-10-27
Version Control Revisions

brymz:


brymz:
My review of #264, #265, #266 is complete and includes issues #269 and #270.

Are there major benefits to using `git` in `RStudio` over the command line?

ethanwhite:
Nice work! Thanks.

ethanwhite:
I think RStudio's system is much less intimidating for first time users. It barely has any features, but I think it's the best way to get new git users started if they're working in R.

PR #276, opened 2015-11-02
Give proper names and docs to size class function

ethanwhite:
In the Version control basic 7 exercise the function to be added had a mix
of information about gc-content and size class due to copy-paste errors.

### DIFF
--- a/exercises/Version-control-basic-7-R.md
+++ b/exercises/Version-control-basic-7-R.md
@@ -34,9 +34,9 @@ Fast-forward
 3\. You should see the new function in your repository.
 
 ```
-gc_content <- function(seq){
-   #Calculate the GC-content for one or more sequences
-   ear_lengths <- ifelse(seq > 10, "large", "small")
+get_size_class <- function(ear_length){
+   # Calculate the size class for one or more earth lengths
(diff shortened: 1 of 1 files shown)


brymz:
Did `ifelse()` just come out with a new release? It's another game changer. I feel like we would have found this over the summer when we were struggling to figure out how R: worked with `if else`.

ethanwhite:
I think `ifelse()` is base R and has probably been around for a long time:
http://www.inside-r.org/r-doc/base/ifelse

brymz:
Is the `ifelse()` function limited in the input of the `yes` and `no`
arguments? I wonder if you could use the function in a higher-order manner
similar to `data.frame()`. It might look like:
```
ifelse(condition,
     transform_something(input, args),
     transform_differently(input, args))
```

For 'Making Choices 2' it seems like we could definitely use the
`ifelse()`, but I'm not sure what the best approach would be for 'Making
Choices 4'.

ethanwhite:
I suspect that the example you show would be fine, but don't know for sure.

I like teaching the basic if-else statements first, but I can definitely see adding  `ifelse()` towards the end of the the assignment.

brymz:
Just something to think about. No strong feelings.

Issue #270, opened 2015-10-27
Choose * or 1. for Version-control-basic exercises


brymz:
The first 5 Version-control-basic exercises use bullets while -6 and -7 use a numbered list. -7 also uses a `\` in its numbered list that appears to alter the alignment of the list.

PR #279, opened 2015-11-03
Versions-control-basics format ordered lists

brymz:
As per #270. I chose to use numbered lists.

### DIFF
--- a/exercises/Version-control-basic-1-R.md
+++ b/exercises/Version-control-basic-1-R.md
@@ -24,11 +24,11 @@ before somehow hitting `Ctrl-s` and overwriting all of your hard word.
 Determined to not let this happen again you've committed to using `git` for
 version control.
 
-Install git for your operating system following the
+Install `git` for your operating system following the
 [setup instructions](http://www.datacarpentry.org/semester-biology/computer-setup/). Then
 create a new project for this assignment in RStudio with the following steps:
 
-a. File -> New Project -> New Directory -> Empty Project
-b. Choose where to put your project
-c. Select `Create a git repository`
-d. If everything worked in the upper right corner of RStudio you should see a `Git` tab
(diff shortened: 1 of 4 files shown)


ethanwhite:
The issue with numbered lists here is that students interpret them as being sub-problems since that's how they are used in most exercises. I recall that we talked about using numbered lists for actual sub-problems and something else for lists of instructions, but don't recall what that 'something else' was.

The use of `1\.` is intentional because otherwise the code blocks will prevent the numbers from properly rendering in order.

brymz:
I also recall a conversation about numbered lists. I believe it was regarding a problem like 'basic-2-R.md' where the tasks are presented without an expectation for what order they are performed, and use `*`. I don't remember a conclusion regarding step-wise problems. I have been numbering step-wise problems, like `Sql-importing-data.md`. I would say that in some ways each step in the exercise is like a sub-problem. It just doesn't have a tangible output. Is the trouble that it isn't always obvious if there's supposed to be output?  

> The use of 1\. is intentional.

Ok. I'll have to go back and make sure the indenting is properly formatted.

brymz:
> > The use of 1. is intentional.
> Ok. I'll have to go back and make sure the indenting is properly formatted. 

I set the code blocks into the list by indenting each line in the code block three spaces. The list should now render in order and maintain consistent indent formatting with the rest of the assignment.

ethanwhite:
> I set the code blocks into the list by indenting each line in the code block three spaces. The list should now render in order and maintain consistent indent formatting with the rest of the assignment.

Wow, that's a new one for me. Thanks!

> Is the trouble that it isn't always obvious if there's supposed to be output? 

Yeah, I've seen students get hung up on this, but since we've been doing it this way already let's go ahead and put this in and we can always revisit later.

Issue #280, opened 2015-11-03
Update answer to R-SQL 1 to incorporate limitation to Controls


ethanwhite:
In f4443bad53027755adfc7d947b29dce5ba657d49 I changed the problem to only apply to the controls so that we needed joins. We need to update the output solutions to match.

brymz:
I'm on it.

brymz:
Thanks for the nudge.

PR #281, opened 2015-11-03
Sql-1-R.txt: update to Control plot values

brymz:
As per #280


brymz:
Merging before review.

Issue #283, opened 2015-11-09
Code solution for Sql-2-R is wrong


ethanwhite:
```
SELECT species_id, sex, AVG(hindfoot_length), AVG(weight)
FROM surveys
WHERE species_id IS NOT NULL AND sex IS NOT NULL
GROUP BY species_id, sex;
```
Yields 49 rows.

ethanwhite:
I think the issue here results from the fact that the code solution loops over all possible species sex combinations instead of just using a GROUP BY. The GROUP BY will be the more common approach.

brymz:
Wow. Of course, `GROUP BY` is the better solution for this question. I am revising the exercise and solution to make the direct connection to `GROUP BY`.

The bigger problem... I constructed this exercise with the goal of integrating a sequence of queries with 'R Fundamentals' [Step 1. make a query; Step 2.: generalize the query in a custom function; Step 3. automate process with a `for` loop]. That no longer happens with the `GROUP BY` solution (*much more elegant*). Do you think that R-SQL integration happens sufficiently throughout the remainder of the course? or do we want to come up with another R-SQL problem / revise 'R-SQL 3' to meet these needs?

ethanwhite:
> Do you think that R-SQL integration happens sufficiently throughout the remainder of the course? or do we want to come up with another R-SQL problem / revise 'R-SQL 3' to meet these needs?

I don't have any strong opinion on this at the moment.

This was closed by #284.

PR #284, opened 2015-11-09
Fix Sql-2-R solution to give correct output

ethanwhite:
Thanks to @bleds22e for catching this and providing the corrected solution.


ethanwhite:
Fixes #283. @brymz can you give this a quick check and then get it in and update the code solution if necessary. Thanks.

ethanwhite:
Going ahead and merging to avoid further confusion. The code solution should be updated when you get the chance.

brymz:
Cool. Thanks @bleds22e for the solution!
Sadly, `for` loops are so 2000s.

PR #287, opened 2015-11-09
Rename and Add output solutions for Put-it-together-R.md

brymz:
Got it done with a fortuitous afternoon nap, but didn't get to check my work. Please, review closely.


brymz:
Merging for class.

PR #295, opened 2015-11-10
Improve wording of Scientific-1-R

ethanwhite:
Also use nrow instead of dim and [1] since it's cleaner

### DIFF
--- a/exercises/Scientific-1-R.md
+++ b/exercises/Scientific-1-R.md
@@ -9,19 +9,18 @@ There were a relatively large number of extinctions of mammalian species
 roughly 10,000 years ago. To help understand why these extinctions
 happened scientists are interested in understanding whether there were
 differences in the body size of those species that went extinct and
-those that did not. Since we're starting to get pretty good at this
-whole programming thing let's stop messing around with made up datasets
-and do some serious analysis.
+those that did not.
 
-Download the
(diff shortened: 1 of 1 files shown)


exercises/Scientific-1-R.md
```diff
@@ -9,19 +9,18 @@ There were a relatively large number of extinctions of mammalian species
...
+To address this question we can use the
+[largest dataset on mammalian body size in the world](http://www.esapubs.org/archive/ecol/E084/094/#data),
+which has data on the mass of recently extinct mammals as well as extant mammals
+(i.e., those that are still alive today). Take a look at the
+[metadata](http://www.esapubs.org/archive/ecol/E084/094/metadata.htm) to
 understand the structure of the data. One key thing to remember is that species
 can occur on more than one continent, and if they do then they will occur more
 than once in this dataset. Also let's ignore species that went extinct in the
-very recent past (designated by the word `"historical"` in the `"status"` column).
+very recent past (designated by the word `"historical"` in the `"status"`
```
brymz: Super picky... `"historical"` gets quotes because it is a character value. `status` does not get quotes because it identifies the column (*which is an object*). I see this sort of markdown styling as part of #87.

ethanwhite:
Good point. Fixed.

Issue #299, opened 2015-11-16
Clean up _config.yml


ethanwhite:
There are a few values that still date to the progbio days or even to Jekyll.

PR #302, opened 2015-11-17
config: rm gems jeckyll-redirect-from; update url

brymz:
Attempt at #299.

I mostly guessed at what needed to be changed in this file by what I could remove (gems) and update (url) without breaking `jekyll serve`.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -7,9 +7,8 @@ permalink:        pretty
 
 # Setup
 title:            Data Carpentry -for- Biologists
-tagline:          'A Jekyll theme'
 description:      'Teaching biologists the tools they need to use computers to do cool science'
-url:              http://www.datacarpentry.org
+url:              http://www.datacarpentry.org/semester-biology
 baseurl:          /semester-biology
 
 author:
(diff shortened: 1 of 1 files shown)


_config.yml
```diff
@@ -9,7 +9,7 @@ permalink:        pretty
 title:            Data Carpentry -for- Biologists
 tagline:          'A Jekyll theme'
 description:      'Teaching biologists the tools they need to use computers to do cool science'
-url:              http://www.progbio.org
+url:              http://www.datacarpentry.org
```
ethanwhite: I think this should be `http://www.datacarpentry.org/semester-biology`

_config.yml
```diff
@@ -9,7 +9,7 @@ permalink:        pretty
 title:            Data Carpentry -for- Biologists
 tagline:          'A Jekyll theme'
```
ethanwhite: We should get rid of this.

ethanwhite:
And you're right about being able to drop the redirect-from.

_config.yml
brymz: Done.

_config.yml
brymz: Done.

brymz:
Ready to go.

Issue #298, opened 2015-11-15
Add a style problem


ethanwhite:
Provide a piece of code with lots of aspects of poor style and the exercise is to clean it up.

brymz:
Are there any poor style examples that you want to be sure are included in the exercise?

ethanwhite:
Off the top of my head the list would include:

* bad variables names (not descriptive, poor capitalization)
* no/inconsistent indenting
* bad spacing (around operators, parentheses
* lack of or bad comments for functions

brymz:
As per https://github.com/datacarpentry/semester-biology/pull/305

PR #305, opened 2015-11-18
add Style-R.md and BADstyle-R.R code

brymz:
As per #298.

### DIFF
--- a/code/magic-8-ballBADstyle-R.R
+++ b/code/magic-8-ballBADstyle-R.R
@@ -1,0 +1,43 @@
+simulate_shake_of_the_magic_8_ball_and_get_answer <- function(shaketime=3, 
+annotation=TRUE, answers_list=answers) { # This function simulates a shake of a Magic 8 Ball.
+cat("Shaking... ")
+Sys.sleep(shaketime/3)
+cat("... ")
+Sys.sleep(shaketime/3)
+cat("Turning... ")
+Sys.sleep(shaketime/3)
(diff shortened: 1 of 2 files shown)


ethanwhite:
My first thought is that it would be nice for this to be a science/biology related example so that the thought process from the code they've been writing (and to the code they will write) transfers a little better. What do you think?

brymz:
I guess I saw this lesson in a similar fashion as the 'Expressions and Variables' and 'Vectors' exercises, where the concepts are fundamental beyond working with data. My goal was to develop a lesson that was a stand alone script (i.e., no download or working directory requirements) that I could guarantee would open and work despite the formatting issues.

Are you opposed to the 8-ball example broadly or just the delivery? Are simulations beyond the scope of 'data science'? I could easily make the `answers` a data file to `read.csv()`. My other thought was to pick an existing exercise where I could poorly format the solution, but then the solution would be out there for grabs.

ethanwhite:
I agree that the concepts go beyond working with data, but that could be said of a lot of what's in this course. Since this is coming at the end of the course there is no need to keep thinks simple like in the early exercises, so it seems like keeping this in the biology data realm would be inherently beneficial.

brymz:
Simulate population time series or sequences.

ethanwhite:
I'm going to go ahead and close this since presumably the eventual solution will be in a new PR. If I'm wrong about that feel free to reopen.

Issue #318, opened 2015-12-11
Cull / integrate extra files.


brymz:
Please, object to any on the list if you disagree:

https://github.com/datacarpentry/semester-biology/blob/gh-pages/style-guide.md
https://github.com/datacarpentry/semester-biology/blob/gh-pages/compbioproj.md

ethanwhite:
Go for it.

brymz:
How about the .PY scripts in `tools/`? Cull or Integrate?

ethanwhite:
Cull

brymz:
How about `atom.xml` in the main directory? It looks like it is linked to `_posts/` directory that we deleted.

brymz:
Maybe `atom.xml` is important. I get this error `jekyll 2.4.0 | Error:  Address already in use - bind(2)`.

ethanwhite:
Let's just leave it then.

Issue #314, opened 2015-12-01
Update and complete material.md.


brymz:
The final materials page should include the assignment readings like the assignments page includes the scheduled assignments. 

- [ ] Add link to `materials.md` in `schedule.md`

Issue #310, opened 2015-12-01
Integrate lecture/demo material


ethanwhite:
There is some good, demo in class style, lecture material. This should be given appropriate YAML headers and should then be automatically integrated into the site with either (or both) a `Classroom Demos` page listing all of the available material or an additional column in the `Assignment Schedule` table.

brymz:
Any objections to combining the `lectures/` and `material/` directories?

ethanwhite:
Nope

PR #324, opened 2016-01-14
Update Lecture Materials

brymz:
As per #310 and #314.

250dd35 - As per #318.


brymz:
This set of commits 1) merges all lecture material and in-house resources to `lectures/` and 2) moves and updates `materials.md` as `lectures/index.md`. 

The next step is to integrate the lecture material into the course, either as a link to an organizing page (i.e., `lectures/index`) or in some way integrated with the schedule. Automating the lecture materials with the schedule is difficult because 1) there are multiple `lecture/` files per week, like exercises and 2) they way I automated the schedule based on assignment `title` (i.e., 'R Assignment 1') would break if I add any more files with matching `title`. The fix might require a new yml field to be added across the board.

brymz:
I restructured the directory, so that exercises : assignments :: materials : lectures. Yaml should be added to all of the most recent course materials and lectures. Please have a look through to make sure I didn't miss anything or put something in the wring place (e.g., `materials/bioconductor.md`). Next, is to have a try at automating the lectures into the schedule.

brymz:
I decided to try a different-but-related title for the `lectures/` yml. It works as long as you don't mind the slightly different naming conventions between the `readings/`, `lectures/`, and `assignments/`. By the precedence, the `lectures/` should share a `title:` with the assignment and a `subtitle:` of `'Lectures'`

Take home message: 
- I've automated the `schedule.md` build to include `lectures/`.
- It's a bit of a work around.
- It's 'ready for review'.

brymz:
I did not update the links on `index.md` to avoid a merge conflict.

brymz:
This is ready for review / merge.

ethanwhite:
Hey @brymz - sorry for the delay on reviewing this. Way too many balls in the air at the moment.

This looks awesome! Great work.

Could you add on more thing before this goes in? Since this is very much a first pass and the lecture materials haven't been gone through to make them very useful to folks other than me I'd like to add a note at the top of each of the assignment lecture pages that says something like:

> Under development

After that we can get this in.

brymz:
Should be ready to go.

> Under development

Added note to the top of `lectures/index.md`, instead of over each of the individual lectures, figuring that should cover it.

Also, updated the schedule as per #341.

ethanwhite:
:+1:

PR #327, opened 2016-01-19
Add Style Problem

brymz:
As per #298.


brymz:
`Style-R.md` is updated. Still need to update contents of `bad-style.R`.

brymz:
Updated code solution `Style-R.R`: https://github.com/weecology/datacarp-semester-biology-exercise-solns/pull/96

brymz:
Ready for review.

exercises/Style-R.md
```diff
@@ -0,0 +1,13 @@
+---
+layout: exercise
+title: Good Style
+subtitle: Modify the Code!
+language: R
+---
+
+Programming computers to do cool science is a major advantage for modern scientists. But, developing research programs that are transparent, collaborative and reproducible is a major advantage for science. All data scientists can contribute to this goal by writing their code in easily readable, well described, and well commented scripts.
+
+This was not the case for the hack that wrote [this script]({{ site.baseurl }}/code/bad-style-R.R) to evaluate [tree composition in a Michigan swamp](http://www.esapubs.org/archive/ecol/E090/251/default.htm). *It works, but it's UGLY!*
```
ethanwhite: Let's change the language here to be lest judgemental. Lots of folks in the class will have written ugly code and we don't want to make them feel bad about that in an unproductive way.

"This was not the case for the hack that wrote [this script]({{ site.baseurl }}/code/bad-style-R.R) to evaluate [tree composition in a Michigan swamp](http://www.esapubs.org/archive/ecol/E090/251/default.htm)"

-> 

"[This script]({{ site.baseurl }}/code/bad-style-R.R), which evaluates the [tree composition in a Michigan swamp](http://www.esapubs.org/archive/ecol/E090/251/default.htm), doesn't follow good style."

'*It works, but it's UGLY!*' -> '*It works, but it's difficult to understand exactly what it is doing.*'

ethanwhite:
Looks great. I've recommended a tone change to avoid discouraging students, but once that's done this is ready to go. Thanks!

brymz:
Thanks for catching that. Changes complete.

Issue #331, opened 2016-02-02
Update _comfig.yml


brymz:
From support@github.com:

The page build completed successfully, but returned the following warning:

You are attempting to use the 'pygments' highlighter, which is currently unsupported on GitHub Pages. Your site will use 'rouge' for highlighting instead. To suppress this warning, change the 'highlighter' value to 'rouge' in your '_config.yml'. For more information, see https://help.github.com/articles/page-build-failed-config-file-error/#fixing-highlighting-errors.

For information on troubleshooting Jekyll see:

  https://help.github.com/articles/using-jekyll-with-pages#troubleshooting

brymz:
I had to run `gem update github-pages` to get the `rouge` highlighter and start using the extended command `bundle exec jekyll serve`to get the site on my `localhost`.

brymz:
Now this...

The page build completed successfully, but returned the following warning:

You are currently using the 'rdiscount' Markdown engine, which will not be supported on GitHub Pages after May 1st. At that time, your site will use 'kramdown' for markdown rendering instead. To suppress this warning, remove the 'markdown' setting in your site's '_config.yml' file and confirm your site renders as expected. For more information, see https://help.github.com/articles/updating-your-markdown-processor-to-kramdown.

For information on troubleshooting Jekyll see:

  https://help.github.com/articles/using-jekyll-with-pages#troubleshooting

If you have any questions you can contact us by replying to this email.

PR #333, opened 2016-02-02
_config.yml: rdiscount to kramdown

brymz:
As per #331.

Looks like `kramdown` does not support code blocks in the same way that `rdiscount` does. This could require a LOT of reformatting. Any ideas? alternates? Seems like GitHub will only be supporting `kramdown`.

Here's the doc for `kramdown`: http://kramdown.gettalong.org/syntax.html . They were nice enough to flag differences between `kramdown` and standard markdown.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -1,5 +1,4 @@
 # Dependencies
-markdown:         kramdown
 highlighter:      rouge
 
 # Permalinks
--- a/_config.yml
+++ b/_config.yml
@@ -1,5 +1,9 @@
 # Dependencies
+markdown:         kramdown
 highlighter:      rouge
+
+kramdown:
+  input:          GFM
 
 # Permalinks
 permalink:        pretty
--- a/_config.yml
+++ b/_config.yml
@@ -1,5 +1,5 @@
 # Dependencies
-markdown:         rdiscount
+markdown:         kramdown
 highlighter:      rouge
 
 # Permalinks
--- a/exercises/Data-analysis-R.md
+++ b/exercises/Data-analysis-R.md
@@ -6,14 +6,14 @@ language: R
 ---
 
 We are interested in understanding the monthly variation in precipitation in
-Gainesville, FL. We'll use some data from the
-[NOAA National Climatic Data Center](http://www.ncdc.noaa.gov/).
+Gainesville, FL. We'll use some data from the [NOAA National Climatic Data 
+Center](http://www.ncdc.noaa.gov/).
 
 Start by creating a `data` directory in the same directory as your homework
 scripts and then downloading [the data]({{ site.baseurl }}/data/gainesville_precip.csv) and saving it to this `data` directory.
 
-Each row of this data file is a year (from 1961-2013) and each column is a month
-(January- December).
+Each row of this data file is a year (from 1961-2013) and each column is a 
+month (January - December).
 
 Rearrange the following program so that it:
 
@@ -21,8 +21,8 @@ Rearrange the following program so that it:
 - Calculates the average precipitation in each month across years
 - Plots the monthly averages as a simple line plot
 
-Finally, add a comment above the code that describes what it does. The comment
-character in R is `#`.
+Finally, add a comment above the code that describes what it does. The 
+comment character in R is `#`.
 
 ```
 plot(monthly_mean_ppt, type = "l", xlab = "Month", ylab = "Mean Precipitation")
--- a/exercises/Expressions-variables-1-R.md
+++ b/exercises/Expressions-variables-1-R.md
@@ -5,9 +5,9 @@ subtitle: Basic Expressions
 language: R
 ---
 
-Think about what value each of the following expressions will return?
-Check your answers using the R Console by typing each expression into
-the console on the line marked `>` and pressing enter.
+Think about what value each of the following expressions will return? Check 
+your answers using the R Console by typing each expression into the console 
+on the line marked `>` and pressing enter.
 
 1. 2 - 10
 2. 3 \* 5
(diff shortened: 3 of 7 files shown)


ethanwhite:
Argh. This could actually be pretty problematic. See https://github.com/datacarpentry/semester-biology/commit/5dec6475348f2480b825c65480b062465c6cac02 for the reason I originally switched to `rdiscount`. Assuming this bug still exists (and it hasn't been fixed in Jekyll from what I can tell) this would break a lot of stuff. Sigh.

First things first, make the change locally and then look around and see what the damage is, both in terms of the code blocks and in terms of the nested lists on assignment pages. Then we'll go from there.

brymz:
Ok. Good news is it looks like nested lists work. But,

- [ ] Code blocks are rendered from line indents (4 spaces?) while \`\`\` are just code lines. 
- [ ] `[ | ]` renders as a table (i.e., list of links in `exercise/index` and `materials/index`).

Unrelated finds:
- [ ] First link in `SQL-practice.md` is broken.
- [ ] R Assinments 9-11 have extra 'Exercises' heading.

ethanwhite:
> Code blocks are rendered from line indents (4 spaces?) while ``` are just code lines or broken. 

I vaguely remember this coming up with Software Carpentry at some point. Does replacing ``` with ~~~ solve the issue?

ethanwhite:
> [ | ] renders as a table (i.e., list of links in exercise/index and materials/index).

Is it escapable using `\`? If not we should be able to design around.

brymz:
Looks like using ~~~ will work in place of \`\`\` for code blocks. Running into errors using `sed` to automate the find/replace process: `sed -i 's/```/~~~/g' Expressions-variables-5-R.md 
sed: 1: "Expressions-variables-5 ...": invalid command code E`. 
Any thoughts?

ethanwhite:
Not off the top of my head. With three separate kinds of special characters getting that sort of thing right can be a nightmare. A text editor or IDE that supports find and replace across files might be worth trying. 

I also noticed that the GitHub blog post on the upgrade suggests that backticked code blocks should be fine:
https://github.com/blog/2100-github-pages-now-faster-and-simpler-with-jekyll-3-0

Are you running Jekyll 3?

brymz:
Yes. I updated last week. 

I also saw that the backticks should be ok, but it's not what I'm seeing.

brymz:
I'm going to pass on this for a week. I've checked my machine is up-to-date and haven't found any documentation that suggests what I've done should be broken. When you (@ethanwhite) get a chance, please run the `_config` branch locally and see if you're getting similar issues.

ethanwhite:
I can't actually seem to get Jekyll 3 and the github-pages gem to play well together. I either get Jekyll 3 running and an error that kramdown isn't installed, even though it is, or I start with a clean install and `bundle install` installs Jekyll 2.4.

brymz:
Yikes. Did you try `bundle update` or `gem cleanup` (suggestions from jekyll docs that didn't work for me)? Seems like github and jekyll are still working some things out. I found at least one comment thread that suggested a bug related to `\`` in a Jekyll 3.0 build. Not sure where to go from here.

brymz:
Making some progress: Updated `_config.yml` as per http://stackoverflow.com/questions/35344518/kramdown-backticks-not-recognized-in-gfm-mode.

Now `####Lesson Objectives` need to be updated in `assignments/` to have a space between the hash and text. But, want to check in about strategies to avoid merge conflict with #338

brymz:
Aside from the required space in headers. This should be ready to go. I'm going to open a new issue to work out merging all of the open projects.

ethanwhite:
I finally got a testing environment working for this (after completely removing ruby from my system and starting from scratch). Looks great overall! This is one of those funny changes where it will look like it probably took 5 minutes, but it was way more work than that.

In addition to the space in the headers I also noticed several problems that didn't seem to have the text reflowing properly. E.g., r-2 problem 4 and r-3 problem (there are several others, these are just examples).

Other than those two things this is ready to go.

brymz:
It looks like what's going on here is that hard wrap is being more rigid and unforgiving of changes in text size. Lines where code blocks extend the length of the line when they are rendered soft wrap, but don't account for the extra hard wrap.

I'm going through to sort out as many of the issues that I can find unless you think there's a behind-the-scenes way to get at this.

brymz:
I've gotten into the bad(?) habit of editing comments as I'm working, instead of writing a new comment each time I think of something to add or modify. Anyway, as I don't think you get email updates when I edit comments, here's a note to check back on the last comment in the thread for the updated details of our formatting problem. I could use your feedback to determine if you think there might be a developmental (behind-the-scenes) solution, or if I should continue going through each of the assignments for formatting edits.

ethanwhite:
Wow, they really did rush this whole Jekyll 3 thing. Markdown parsing is supposed to convert hardwraps in the .md documents into softwraps in the html. Apparently they broke this and recently unbroke it:
https://github.com/jekyll/jekyll/issues/4471

I just updated Jekyll and everything is working fine now. Sigh.

Was that the last thing to deal with this in PR?

brymz:
Yep. Fingers crossed for no more surprises.

ethanwhite:
OK, two things here:

1. We don't want the extra changes in e33861f. Since the history for the change is already a little messy (3 commits to get it right). Let's just start a new branch, make the changes to `_config.yml` in a single commit, and submit a new PR.
2. Github is still [running on Jekyll 3.0.3](https://pages.github.com/versions/), which from a quick glance doesn't have this fix in place. So once you submit the new PR it may be worth sitting on this for a little bit while they catch up.

Issue #338, opened 2016-02-06
Condense weeks 3-5 into 2 weeks


ethanwhite:
In the last run of the course weeks 3-5 felt a little thin to me in total, so I think we can combine them into two weeks. My specific thoughts (which are certainly open to discussion) involve:

1. The first week combines the current first week and the first two exercises from the second week. We should probably include a super simple data frame problem, so the 3rd problem from the 2nd week, but removing some of the more complex bits, at least including the use of `subset`.
2. The second week is then more advanced data frame stuff + dplyr. Since we're already showing them dplyr we can probably skip the part of the 4th problem from the current 2nd week that has them manually export data and then load it into R. Maybe move this problem to be the last problem of the new 2nd week and have them do the query from R using dplyr instead of doing the manual export.

What do you think? Is this still a reasonable amount of material in each of these two weeks for beginners? Any suggested changes to the plan for the reorganization? Other thoughts?

brymz:
Are we going to keep the 'SQL Practice' exercise?

ethanwhite:
Sure, I think a little basic revisiting of SQL on it's own isn't a bad thing, but I'm open to being convinced otherwise.

brymz:
> I'm open to being convinced 

My concern is that we lose adequate time and brain power for new material by forcing it in with the condensed Intro-R weeks. If anything, I would put it right at the beginning of week 3 before moving on to R. Or leave it out and gauge the next run of class, inserting it if the need arises.

'Scientific 0' is the `dplyr` version of 'Data Frames 1'. Do you want to maintain the `subset()` and `object['colum_name']` subsetting material or just jump from 'here's what a data frame looks like'  to `dplyr?`

ethanwhite:
OK, I think that's a good argument. How about moving it to the beginning of the Data Visualization assignment?

I'm leaning towards jumping straight to `dplyr`. What do you think?

brymz:
'SQL Practice' to Data Visualization assignment, aye.

Straight to `dplyr` seems the DC model and helps clean the merge of weeks. If anything, detailed subseting operations can be introduced with the R Fundamentals assignments.

ethanwhite:
:+1:

PR #339, opened 2016-02-09
Combine Intro-R weeks

brymz:
As per #338.

Weeks are condensed largely by removing detailed R{base} subsetting functions. 'Data Frames 2' has been removed from the course material, with relevant content moved to the remaining `dplyr` lessons. @ethanwhite, have a look to make sure you like the structure and then I will go ahead and update the code and output solutions.


brymz:
This is ready for review and merge.

One thing I noticed in 'Dplyr Using Databases': The last few 'queries' return `tbl_df` that are `<derived table>` and do not return any `nrow` value. What's more, I couldn't figure any way to interact with the values in these tables. Probably not a big issue here as the exercise questions just ask for tables, but students might stumble on if they are messing around. My only explanation is that it is an `RSQLite` issue as I've never run into this before with `dplyr`.

exercises/Data-frames-1-R.md
```diff
@@ -15,14 +15,7 @@ their volumes (`l * w * h`). You could do this using a spreadsheet, but the
 project that you are working on is going to be generating lots of these files so
 you decide to write a program to automate the process.
 
-Download the data, use `read.csv()` to import it into R, and then print out:
+Download the data, use `read.csv()` to import it into R, and then use the `$` operator to print out:
```
ethanwhite: What is the use of `$` for here? I think I was envisioning just printing out the entire data frame.

ethanwhite: Never mind. Coffee kicking in now.

ethanwhite:
Looks great! The only thing I see is the need to deal with the fact that Assignment r-3 is now empty. We should cascade the number moves so that 4 becomes 3, 5 becomes 4, and either 8 disappears or the spatial lesson gets inserted into the appropriate location even though it will be blank temporarily (as an aside, this sort of thing is why I'm still not overly thrilled about using numbers instead of names for assignments, especially since some folks might not want to teach a week on spatial analyses, but we should save that for another issue).

I'm guessing this may have some merge implications, which we can talk about later today.

brymz:
Last week, we discussed merging this before I made further changes to help avoid merge conflicts. Any last minute objections, @ethanwhite?

ethanwhite:
Nope. Just too many things going on. :scream_cat:

PR #346, opened 2016-03-24
_config.yml: update markdown & kramdown input

brymz:
As per #333.

Considering delayed merge until GitHub updates Jekyll.


ethanwhite:
Given that this switch is coming up soon, the next time you have some time to work why don't you go ahead and merge this and then do a run through of the site and open new issues for any problems it triggers.

Issue #345, opened 2016-03-01
Fix direct-to-file figshare links


ethanwhite:
It looks like figshare changed it's direct-to-file link structure and broke any of our links that pointed directly to files in the Portal Project Teaching Database.

Here's the one example of a fix so far in a data carpentry lesson:
https://github.com/datacarpentry/spreadsheet-ecology-lesson/pull/41

PR #350, opened 2016-03-24
Fix Figshare links

brymz:
As per #345


ethanwhite:
May I just say that nicely delineate PRs and clear commit/PR messages make it really easy to move through reviews quickly. It's really nice tradecraft.

PR #359, opened 2016-04-08
Add R-SQL lecture

### LINKED ISSUE #355: Add R - SQL Lecture to materials/; Add links to  r-sql.md

ethanwhite:
Fixes #355

### DIFF
--- a/lectures/sql-R.md
+++ b/lectures/sql-R.md
@@ -1,0 +1,54 @@
+# R-SQL
+
+* Already seen how to connect to database using `dplyr`
+* What if we want to do this outside of a `dplyr` context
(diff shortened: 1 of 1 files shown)


brymz:
To match the current site structure, this file needs to be moved to `materials/` and a lectures page set up in it's current place with a link to the file. I'm going to merge this now and make the changes in another PR. BUT, this issue brings up something I've been thinking about before. 

Some lectures are a single file and some are broken up into multiple files. I've set up the site structure so that multiple files from `materials/` are organized in a `lectures/` file. This causes 'extra clicks' in cases where there is only one file in `materials/`and therefore only one link at the `lectures/` page. Something to think about for the future. You can open an issue if you think it warrants revisiting.

PR #362, opened 2016-04-15
Move R-sql lecture to materials and add link to lecture page

brymz:
Follow-up on #359.


brymz:
Some minor troubleshooting problems that will cause extra commits. Closing and reopening fresh PR.

PR #379, opened 2016-04-22
Update README

brymz:
As per #364.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,19 +1,21 @@
 # Data Carpentry for Biologists - Semester Course
 
-This is **a [forkable](https://help.github.com/articles/fork-a-repo/) set of teaching materials** for teaching biologists how to work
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -98,18 +100,18 @@ and the standard YAML for a new exercise would look like this:
 ```
 ---
 layout: exercise
-title: Name of exercise
-subtitle: Short description of the exercise
-language: R
+title: Name_of_exercise
```
ethanwhite: I actually prefer the spaces here and on the next line because it makes clear that these values can include spaces.

ethanwhite:
Looks good overall. I don't like the use of bolding inside of sentences. I tend to think that if one needs that sort of thing it's a sign that the text should be more concise instead. I also made an in-line suggestion about spacing.

brymz:
Bold text and `_` removed. Ready to go.

Issue #367, opened 2016-04-15
Add exercise title to assignment.html


brymz:
Reference for formatting: http://kramdown.gettalong.org/quickref.html

PR #380, opened 2016-04-22
WIP: Reconcile exercise structure

brymz:
The current exercise structure uses `title` and `subtitle` in some opposition. `title` is the file name and key used to link exercises to assignments and back to related exercises (i.e., "This is a follow-up"). `subtitle` is the descriptive tag used in the list of exercises in the assignment. So, when you scroll through an assignment you see a follow-up exercise that has a name you've not seen before unless you've checked out the `exercise/index` or `github.com`. I see a couple of ways to resolve the issue:

1) Make the `title` more visible. This patch adds the exercise title to the `assignments` exercise list, as per #367. With this change, we will want to revise some of the existing `title` & `subtitle`.

2) Update the "follow-up" text with `subtitle`. With this change, some `subtitle` might need to be revised. We might also want to consider changing the format of `exercise/index` and `exercise.html`.

Now that I've seen the start of option 1, I might be leaning towards option 2. Option 2 would partition the need to know the `title` to developers. Students would only see the `subtitle`. 

Any thoughts?

### DIFF
--- a/_includes/assignment.html
+++ b/_includes/assignment.html
@@ -3,7 +3,7 @@
 {% for exercise in page.exercises %}
   {% for otherpage in site.pages %}
     {% if exercise == otherpage.title and page.language == otherpage.language %}
-      <li><h4>{{ otherpage.subtitle | prepend: "--" | append: "--" }}</h4>
+      <li><h4>{{ otherpage.title | prepend: "-- " | append: ":" }} {{ otherpage.subtitle | append: " --" }}</h4>
         {{ otherpage.content }}
         {% capture output_file %}{{ otherpage.url | remove: 'exercises' | remove: '/' | prepend: '/solutions/' }}{% endcapture %}
         {% for solution in site.static_files %}


ethanwhite:
I think we should set up some time to chat about this. I'm have a gap in my schedule tomorrow afternoon at 1 pm Eastern if that happens to work.

brymz:
Works for me. See you tomorrow (Tues) 1p Eastern, 12p central.

brymz:
Our goal is to develop a website structure that is intuitive for both students to view and instructors to develop. Right now, the `title`, `subtitle`, and exercise file name overlaps in organizational and descriptive function. The challenge is to include both organization and description in the naming conventions. The solution will likely require another major overhaul of the site structure with new yaml tags and file names. I am favoring the idea to keep two yaml tags (`topic` and `title`) and develop a file name that uses both of these pieces. I think the key will be dropping the numbering system as we did for the `assignments`. I will open an issue with the intention to digest our conversation and outline a solution.

PR #393, opened 2016-04-28
WIP: Update Site Navigation

brymz:
As per #384.

PR #394, opened 2016-05-03
Update Site Navigation

brymz:
Fix merge conflict in #393. *All I did was merge the `nav` branch with an updated `HEAD`. I'm not sure how that would resolve a merge conflict. There were no files flagged for conflict.* 

The PR develops a hierarchical navigation strategy that limits the number of top-level navigation links on the sidebar and generates a second level of link pages organized in `nav/`. It is ready for review.


ethanwhite:
Looks great in general. A couple of things:

1. The LICENSE file needs to stay in the root directory because that's the standard.
2. `main` doesn't seem very intuitive as the location that I'd look for things like the syllabus. We should either come up with a different name or put those 4 files back in root.

brymz:
I have moved the `LICENSE.md` and `main/` files back to the root directory and fixed the affected links. Should be ready to merge.

Intuitively finding the course materials in `main/` was a concern of mine as well. My alternate suggestions `home/` and `start/` suffer from the same problem. I was excited to get most of the .MD files out of the root directory, but we want to be sure important documents can be found easily. Unless you have a suggestion for a better place to put those `main/` files we can just leave them in the root. I've worked out automation of file moves and link fixes, so any new decision should be easy to make happen anytime.

PR #401, opened 2016-05-31
Restructure `nav/` links to <dl>; Add link icons 

brymz:
This is a minor patch for flavor. It is a visual change to the navigation structure of the site. You will want to run the PR locally to see the update.

### DIFF
--- a/_includes/sidebar.html
+++ b/_includes/sidebar.html
@@ -9,10 +9,10 @@
       <p class="lead">{{ site.description }}</p>
     </div>
     <nav class="sidebar-nav">
-      <a class="sidebar-nav-item{% if page.url == '/nav/getting-started/' %} active{% endif %}" href="{{ site.baseurl }}/nav/getting-started">Getting Started</a>
-      <a class="sidebar-nav-item{% if page.url == '/nav/course-materials/' %} active{% endif %}" href="{{ site.baseurl }}/nav/course-materials">Course Materials</a>
-      <a class="sidebar-nav-item{% if page.url == '/schedule/' %} active{% endif %}" href="{{ site.baseurl }}/schedule">Schedule</a>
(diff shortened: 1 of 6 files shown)


ethanwhite:
I really like this and think it makes the site very visually appealing. Awesome work!

We can accomplish at least some of this without dropping to html because `kramdown` supports `dl`. Here's the `Getting started` page in markdown:
```
Self-Guided Students
: <a href="{{ site.baseurl}}/START-for-self-guided-students">
  <i class="fa fa-play-circle"></i> Start Here</a>
  
Semester-long Course Students
: <a href="{{ site.baseurl}}/computer-setup">
  <i class="fa fa-download"></i> Setup</a>
: <a href="{{ site.baseurl}}/syllabus">
  <i class="fa fa-file-text-o"></i> Syllabus</a>
: <a href="{{ site.baseurl}}/schedule">
  <i class="fa fa-calendar"></i> Schedule</a>

For Instructors
: <a href="{{ site.baseurl}}/nav/course-materials">
  <i class="fa fa-university"></i> Course Materials</a>
: <a href="{{ site.baseurl}}/docs">
  <i class="fa fa-question-circle"></i> Course Development Documentation</a>
```

What I haven't been able to figure out how to do yet is to use these without the definition line (e.g., `For Instructors`) as you've done on some of the pages that don't need headers. Here are the two places I've been looking so far to understand this:

* http://kramdown.gettalong.org/syntax.html#lists
* http://makandracards.com/makandra/6683-markdown-kramdown-examples

I'll let you take a crack at getting this all converted to this syntax and think that once you do this is definitely going in.

brymz:
The kramdown formatting is definitely much cleaner. I wasn't able to find any documentation that allowed for a blank line for the 'term', but decided that the extra indentation on those pages was not important.

So, I removed the `<dl>` tags from all of the pages and left the kramdown `:` where there were 'terms'.

Issue #400, opened 2016-05-31
Consider converting content folders (i.e., assignments/, exercises/) to collections.


brymz:
- [Jekyll collections documentation](https://jekyllrb.com/docs/collections/)
- Organizing `assignments/` and `exercises/` as collections might resolve the need to have `assignments/` files in lower case and `exercises/` capitalized.
- Including `lectures/` and `materials/` as collections might allow me to remove the `element:` yaml tag.

brymz:
From @ethanwhite (#396): "Nice find! It would be great to try this out since I hate that hack."

PR #404, opened 2016-05-31
WIP: Restucture course materials as collections

brymz:
As per #400 

- exercises in the assignments do not render to markdown
- check `element:` yaml tag
- check default `layout:` yaml tag


brymz:
I have not been able to work out the rendering issues in the `assignments/`pages. It looks like this problem is similar when trying to include content from `posts` in a file in a collection (https://github.com/jekyll/jekyll/issues/3864).

At the moment, it seems the rendering issues driven by processing ordering would not be resolved by using collections.

ethanwhite:
OK, no problem. I definitely spent *a lot* of time trying to figure out how to get this to work sanely without success, so I'm sadly not surprised. Feel free to go ahead and close this if it's dead ended for the moment.

Issue #377, opened 2016-04-22
Generate Course Website Template


brymz:
As per 50173dd9b54966eb8e6e35b4e7ebb5d7cc8d3272.

"some materials that are specific to a given course (e.g., the syllabus) will be generalized to templates instead of being specific to the current course"

- [x] syllabus
- [x] config set-up
...

brymz:
What is the best way to store the course website template? separate git repo? `template/`?

brymz:
Changes to also include back in the course website:

- _config.yml
- [ ] add `email:`
- 404.html
- [ ] site.github.repo
- [ ] site.email
- index.md
- [ ] rm extra line

PR #409, opened 2016-06-06
Minor revisions to main repo found during template development.

brymz:
As per #377.

### DIFF
--- a/404.html
+++ b/404.html
@@ -10,8 +10,8 @@ permalink: 404.html
     <ul>
       <li>Navigate back to the closest page index from where you are by removing the last piece of the url after the last /.</li>
       <li><a href="{{ site.baseurl }}">Head back home</a> and try to directly navigate to the page.</li>
-      <li>Check out the <a href="https://github.com/datacarpentry/semester-biology/">GitHub repository</a> that builds this site.</li>
-      <li><a href="https://github.com/datacarpentry/semester-biology/issues/new">Open a new issue</a>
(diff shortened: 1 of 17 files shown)


brymz:
This is ready for review. I had considered revising `syllabus.md` to `site.baseurl` and `site.email`, but thought you might want to print the syllabus from its text file before rendering.

brymz:
I revised `syllabus.md` to include yaml `site` and `page` tags. Now, all of the class-specific information is stored in the front matter.

ethanwhite:
Looks really good overall. I love how you handled generalizing the syllabus.

The one thing I noticed is that `START-for-self-guided-students.md `, `audience.md`, and `contributing.md` now point to the course site instead of the central site and I'm wondering what we actually want there. My initial reaction is that feedback and contributions are best directed to the main repo, but I'm certainly open to discussion on this.

brymz:
Thanks!

It was also unclear to me what was the best approach for repo links. In my mind, feedback can be described as relating to course material or site functionality. For the main repo, feedback goes to the same place. But maybe that shouldn't be the case for forked repos. I think `START-for-...-students.md` can always be described as course material and directed to `{{ site.github.repo }}`. Perhaps, `audience.md` and `contributing.md` could use a revision (and maybe an `{% if %}`) to separate out the two types of feedback. 

There might already be a merge conflict with #417. I would prefer to get this in and then resolve conflicts and address the repo links issue there.

ethanwhite:
Sounds good. Done.

Issue #421, opened 2016-06-15
Lectures 'Under Development' - Feedback


brymz:
No fix
- [x] Consider identifying language of code blocks
- [x] Decide on (aside text) vs. (*aside text*) vs. ( *aside text* ) and make necessary edits
- [x] `_` to `-` in file names. (#429)

Fixed
- [x] `Scientific-shrub-volume-3-R.md`: Item 10 out of list (#423)
- [x] `Version-control-basics-adding-a-remote-R.md`: fix nested list (#423)
- [x] `Graphing-adult-size-vs-newborn-size-1-R.md`: Item 5 'go' -> 'do' (#423)
- [x] `assignments/sql-intro.md`: extra `'Creating Tables'` (#424)
- [x] `Version-control-set-up-git-R.md` duplicate `title` (#425)
- [x] Better introduce `Basic-data-management-review-R.md` to `r-git.md` (#426)
- [x] Add `exercises/` link to `assignments/` (#427)
- [x] `.IPYNB` to `materials/` (#430)
- [x] Check for `.` after follow-up links ( i.e. `'Shrub Volume 3'`) (#432)

PR #430, opened 2016-07-07
mv .ipynb to materials/; mv python lecture data to data/

brymz:
As per #421.

I think that the `.IPYNB` that link to the relocated data files will require edits of the file path.


ethanwhite:
Let's go ahead and handle checking and fixing any need to update data file links as part of this PR so that things don't get out of sync.

brymz:
I updated the data file links... and some other things happened.

- I updated my machine to Python 3 and installed iPython Notebooks. The new software did not play nice with the out-dated versions of `ipynb`. I was not able to figure out how to connect the `ipynb` to a working `kernel`. This meant that I could not run any of the code cells for testing or get new output cells.  I tested changes to the data file paths and code with a `New` `ipynb`.
- Opening the old `ipynb` in the new software automatically updated the `ipynb` to v4.
- I updated `print()` to Python3 for the few files with broken file paths. 

This PR could be ready to merge... or given the status update, could launch an effort to update all of the Python material.

ethanwhite:
The notebooks work fine for me in a current version of Jupyter. I open them, they get converted, and then the execute as expected.

How did you install the notebook? It seems like there's a problem with your install that we should get fixed so that you can check all of these if we're going to keep them.

brymz:
I installed Anaconda and then `jupyter notebook`.

ethanwhite:
1. Mac or PC?
2. What does running `which python` and `which jupyter` from the command line generate?

brymz:
MAC OS X Yosemite V 10.10.5
`which python` - `//anaconda/bin/python`
`which jupyter` - `//anaconda/bin/jupyter`

PR #440, opened 2016-08-10
Minor copy-editing

gvwilson:
Tidying up a bit while reading - the notes look to be in *excellent* shape.  One thing I did see is:

~~~
exercises/Tests-fix-the-code-3-R.md:`test_that("get_gc_content() works",`
exercises/Tests-fix-the-code-3-R.md:`test_that("get_size_class() works",`
~~~

Is this a formatting glitch?

### DIFF
--- a/exercises/Version-control-basic-set-up-git-R.md
+++ b/exercises/Version-control-basic-set-up-git-R.md
@@ -1,7 +1,7 @@
 ---
 layout: exercise
 topic: Version Control Basic
-title: Set-up Git
+title: Set Up Git
 language: R
 ---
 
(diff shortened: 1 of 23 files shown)


exercises/Version-control-basic-set-up-git-R.md
```diff
@@ -1,7 +1,7 @@
 ---
 layout: exercise
 topic: Version Control Basic
-title: Set-up Git
+title: Set Up Git
```
brymz: Please, update `exercises:` in `assignments/r-git.md` to match this revision.

brymz:
Thanks for the detailed review and the positive comments!

> Is this a formatting glitch?

I don't think so. Could you be more specific about what you see?

I noticed a different error in the code solution file, but the code runs well with that update. Line 4 should be: `source("Debugging-testEdits-R.R")`.

PR #453, opened 2016-08-19
Update Schedule for Fall 2016 course at UF

ethanwhite:


materials/knitr.md
```diff
@@ -1,6 +1,6 @@
 ---
 layout: page
-element: notes
+element: lecture
```
brymz: Glad to see that this will work. But in the interest of consistency, I recommend that you revise the `lectures/r-extras.md` `title:` to `Knitr` and return this file to `element: note`. I'm not very excited about this organizational structure right now (esp., when there is only one `note` for a `lecture`), but it is how I get multiple `note` to work together.

ethanwhite:
Replaced by #456.

PR #457, opened 2016-08-22
Add an embedded form for getting in class feedback on exercises.

ethanwhite:
Embeds the form as an iframe on the site. The current implementation uses a
Google Form, but by switching out the url in the config file this should be able
to support most systems.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -14,7 +14,7 @@ description:      'Teaching the tools to get computers to do cool science'
 url:              http://www.datacarpentry.org/semester-biology
 baseurl:          /semester-biology
 email:            datacarpentrysemester@weecology.org
-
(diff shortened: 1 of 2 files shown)


ethanwhite:
Not sure of the best place to link this. Possible on the main page after `schedule`, but would need to figure out a font awesome icon.

brymz:
* Are you looking for assignment or exercise level feedback?
* Do we want students to b able to identify which assignment/exercise they are working on?

ethanwhite:
This is for live in-class feedback. I'll use the same poll repeatedly just to see how things are going on a the single exercises that the students work on after the initial intro to a concept. I'll be deleting the responses after every use.

brymz:
Ok. Link on the `sidebar`?

ethanwhite:
Sounds good. Assuming this is otherwise good why don't you merge this, track down a good icon, and add the link to the sidebar. Thanks.

brymz:
Aye, sir.

PR #462, opened 2016-08-23
Add code for first day demo

ethanwhite:
I end the first day of class by demoing some of the things the students will
be able to do by the end of the course. This code downloads BBS and Bioclim
data and combines them to look at the influence of precipitation on species
richness.

### DIFF
--- a/code/first_day_demo.R
+++ b/code/first_day_demo.R
@@ -1,0 +1,33 @@
+library(dplyr)
+library(ggplot2)
+library(raster)
+library(ecoretriever)
+
+if (!file.exists('bbs.sqlite')){
+  ecoretriever::install('BBS', 'sqlite', 'bbs.sqlite')
+}
+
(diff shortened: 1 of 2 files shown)


brymz:
Looks good.

I'm curious about this example because there are a number of things that we won't directly teach students: `ecoretriever`, `dplyr::collect()`, `SpatialPointsDataFrame()`

ethanwhite:
> I'm curious about this example because there are a number of things that we won't directly teach students: ecoretriever, dplyr::collect(), SpatialPointsDataFrame()

We should probably add `collect()`. I noted that we won't teach the retriever but it allowed me to get to using some big scale data. `SpatialPointsDataFrame` was just how I quickly figured out that morning to get the routes data connected to the bioclim data. I'd be happy to have that updated to match how we're doing it in the spatial lesson (or vice versa).

Issue #448, opened 2016-08-17
Alternative reading material for dplyr


ethanwhite:
http://r4ds.had.co.nz/transform.html

See also #446 #447.

Issue #447, opened 2016-08-17
Alternative reading material for tidyr


ethanwhite:
Same source as #446. Hadley is writing a new book and it has a lot of useful well done material for his ecosystem:

http://r4ds.had.co.nz/tidy-data.html

Issue #446, opened 2016-08-17
Alternative reading material for data viz


ethanwhite:
It's more comprehensive than what we have now and nicely put together:
http://r4ds.had.co.nz/data-visualisation.html

PR #463, opened 2016-08-24
Add 'R for Data Science' to Readings

brymz:
As per #446, #447, & #448

### DIFF
--- a/readings/R-data.md
+++ b/readings/R-data.md
@@ -11,5 +11,7 @@ language: R
 
 * Readings
 
-  * [R for Data Science - Data transformation](http://r4ds.had.co.nz/transform.html)
-  * Optional References: [`dplyr` vignette](https://cran.rstudio.com/web/packages/dplyr/vignettes/introduction.html) \| [Analyzing data with `dplyr`](http://www.datacarpentry.org/R-ecology-lesson/04-dplyr.html)
+  * [`dplyr` vignette](https://cran.rstudio.com/web/packages/dplyr/vignettes/introduction.html)
+  * *Optional Resources*: 
+    * [Analyzing data with `dplyr`](http://www.datacarpentry.org/R-ecology-lesson/04-dplyr.html)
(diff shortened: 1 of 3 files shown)


ethanwhite:
Did you take a look at those alternatives and think they were better? That was my first impression but I didn't spend a lot of time comparing the options.

brymz:
> Did you think those alternatives were better?

If by 'better' you mean 'more in-depth' then definitely. He does a really nice job of breaking up the concepts and introducing the package utility in a digestible manner. I learned a few things. There is definitely a bit more than we cover in our exercises. Our exercises are like word problems to his basic exercises. In the case of `dplyr` and `ggplot2`, I left the vignette references because they are more direct in their delivery of the function uses and might be easier for students to return to.

ethanwhite:
I don't just mean 'more in-depth' because that could also mean too long or too detailed. So I guess what I'm asking for is your thoughts on whether these or the existing readings are a better introduction to the kinds of things (but not necessarily the specific functions) that we want the students to know. If you were taking the course and were to read one set of materials in each case, which one would you want it to be.

brymz:
I think the 'R4DS' reading for `ggplot2` is a great introduction to the Hadleyverse. He writes it as the first coding chapter and I think it is the best resource we have for data visualization. As for `dplyr` and `tidyr`, I think the vignettes (that we used earlier) are sufficient and succinct resources for our students. Though I did learn interesting things by reading the 'R4DS' chapters, I think they are a bit verbose for our needs.

ethanwhite:
OK, I've looked over things again and agree with you. Let's move this PR to switch out the `ggplot2` reading, keep the original readings for `dplyr` and `tidyr`, and maybe add a supplemental readings section to those readings pages and include the associated R4DS chapters (or add them parenthetically as supplemental or something along those lines).

brymz:
> Let's move this PR to switch out the `ggplot2` reading, keep the original readings for `dplyr` and `tidyr`, and ...

I went with '*Optional Resources*'. Ready for review and merge.

PR #467, opened 2016-08-25
Add lecture materials on aggregation in SQL

ethanwhite:


materials/sql-aggregation.md
```diff
@@ -0,0 +1,47 @@
+---
+layout: page
+element: notes
+title: Basic Queries
```
brymz: Aggregation?

materials/sql-aggregation.md
```diff
@@ -0,0 +1,47 @@
...
+```
+SELECT species_id, MIN(weight), MAX(weight), AVG(weight)
+FROM surveys
+GROUP BY species_id;
+```
+
+* We can group by multiple columns as well.
+
+```
+SELECT species_id, plot_id MIN(weight), MAX(weight), AVG(weight)
```
brymz: `..., plot_id, MIN(weight), ...`

brymz:
Looks good. Just a typo and a missed(?) `title:`. 

Curious, does SQL automatically remove missing values for aggregation functions? Something I was thinking might want to be added as a note.

ethanwhite:
Fixed.

Issue #482, opened 2016-08-31
Avoid displaying problem numbers on assignments


ethanwhite:
This week the following happened:

1. I decided to add a new SQL exercise on joins that was simpler than all of our existing exercises. The idea was to give the students something easy to get started on.
2. Since there were already Joins 1-5 I named this exercise Joins 0, thought I think Joins 6 would have turned out to be just as confusing.
3. I made this exercise number 1 in the assignment.
4. I asked the students in class to work on the "first joins exercise". A decent fraction of them ended up working on Joins 1 instead of Joins 0.

This illustrates a general issue with our current system of numbering exercises and displaying those numbers in the assignments: If someone wants to change the order of exercises, or add or remove one, having the numbers makes thing confusing. E.g., "Why is Joins 3 missing?" or "Should Joins 2 be before Joins 3?".

One simple approach to avoiding this would be to stop displaying the title in the assignments, but I'm open to other suggestions.

brymz:
> a general issue with [...] numbering exercises and displaying those numbers.

I totally agree.

> ... our current system ...

I disagree. I would argue that using numbers to organize assignments has not been 'our current system' since the major restructure I did over the summer with #395. 

As you have noted, I think Joins remains the glaring exception to this and I would be glad to make an update. My hesitancy is to choose a set of names without a good understanding of what each Joins problem is out to accomplish. Alternately, the problems can be named by their named output table.

Numbers remain in a few other cases where exercises share a theme and linked in a 'follow-up' sequence.

I can get the Joins restructured today if your say, 'Go'.

ethanwhite:
> I disagree. I would argue that using numbers to organize assignments has not been 'our current system' since the major restructure I did over the summer with #395. 

Ah yes, you're totally right. Sorry. Don't know if you're an Ani Difranco fan, but I regularly describe myself as having a "little plastic castle" memory.

Concepts:

JOIN 0: Simple join
JOIN 1: Multi-table join
JOIN 2: Multi-table join with filtering
JOIN 3: Multi-table join with filtering (might be nice to add something else here, maybe a sort on weight)
JOIN 4: Join with aggregation

Take it away.

PR #483, opened 2016-08-31
Update JOIN file names and content

brymz:
As per #481 & #482


ethanwhite:
Thanks @brymz!

brymz:
This is ready to go.

One comment / clarification on `Advanced-queries-detailed-join.md`:

> Be sure to choose only rodents and exclude individuals that have not been identified to genus.

My solution to this statement is `WHERE taxa='Rodent' AND  species_id!='UR'`. 

- [ ] That's ok. That's what I expected.
- [ ] Add clarification to exercise to encourage this solution.
- [ ] Add other clarification (like `WHERE genus IS NOT NULL`) and revise solution accordingly.
- [ ] Remove this condition from the exercise and solution.

ethanwhite:
I agree it's unclear and don't have strong feelings about the right solution. I guess in hindsight I'd actually prefer it to be "identified to species" with a `species_id <> 'sp.'`. If we change that I'll need to get them anyone who's finished with it to update their answer, but in all likelihood most of them won't have done it yet anyway.

brymz:
I'm unfamiliar with the `<>`.

ethanwhite:
`<>` is technically the SQL standard for not equals. See https://en.wikipedia.org/wiki/SQL. That said, almost all databases also accept `!=`, and since that's what it is in R there's probably an argument for teaching `!=` instead.

brymz:
That's a game changer. Needless to say all of my solutions use `!=`. We might consider putting something about this in `materials/sql-basic-queries/`. I'll get started on the exercise and solution revision.

ethanwhite:
I generally show them both, so it's certainly not a big deal to stay with `!=`. Agree that we should explicitly introduce `not equals` in `sql-basic-queries`. What do you think about mentioning both but stating that we'll focus on `!=` since it's consistent with other languages.

Issue #488, opened 2016-09-06
Add new Fall 2016 `materials/` to `materials/index`


brymz:

brymz:
I wrote a `bash` test to find the handful of new files that you've added and was surprised by the lot that was returned to me. So,

Do we want all of the files in `materials/` on `materials/index`?
Should I take the opportunity to update `IPYNB` links from `ethanwhite/progbio` to `datacarpentry/semester-biology`?

ethanwhite:
* Can you post the list of files in `materials`.
* Image files that aren't referenced should probably just be removed
* At this point I think the IPYNB's are probably pretty rusty, so my initial thought is either not publicize them or even remove them. Thoughts?

brymz:
> Can you post the list of files in `materials`?

There are 143 of them, which is why I was thinking the best way to list them for review would be organized on `materials/index`. Otherwise, where do you recommend I post the list?

> Image files that aren't referenced should probably just be removed.

I mostly agree. The handful I would consider holding on to are `databases-basic-table-structure.png` and `databases-crosstab-table.png`. They aren't used in the current version of `database-structure.md`, but could be useful in the future. (*A side note, I was surprised to see you remove this material from your SQL lectures. Perhaps you found you covered the material actively in `tidy-data.md`, but it seems glossed over from the notes.*)

> I think the IPYNB's are probably pretty rusty

That's probably true compared to the progress we have made with the R materials. I think most of them are still pretty strong, perhaps better than some of the accompanying `MD` notes and they are what remains of any Python material. Again, I think organizing them will help us make a thorough investigation of the state of things. I do support removing files that are confusing or incomplete.

You bring up an interesting point about keeping files while not publicizing them. Does that mean we intend to get back to them to make them publishable? They were published at one time as 'Programming for Biologists'. Otherwise, what is their purpose unpublished but present in the directory?

PR #489, opened 2016-09-07
Clean materials/

brymz:
545d76f as per #488 title.

Remaining work as per #488 comments.


brymz:
So, what's happening here?

#488 reminded me that `materials/` was the remaining location that had not fully integrated and updated the 'Programming for Biologists' materials with 'Data Carpentry for Biologists'. I am going through all of the materials to: 

* add links to the `materials/index`
* make sure front matter is sufficient for rendering  
* move script files to `code/`
* remove blatantly out-dated files

My hope is that the remaining files will be presentable and accessible from `materials/index` for you (@ethanwhite) to quickly go through and propose other deletions or merges.

brymz:
Ok. This is ready for a review.

There is definitely some room to shed some extraneous/duplicate lecture material. I have tried to group those files to make it easy to compare and choose. Likely, the greatest set of dismissals will come from the **Extra Notes** sections that include materials that didn't line up with the previous **Course Notes** structure. I do think some of these materials remain useful, but definitely for instructors or intermediate-expert users that are looking for a reference.

If we were to set this aside for a while, the pieces of immediate concern for #488 are:

* 545d76f
* fix 'Project Decomposition' and 'Version Control' links (i.e., `decomposition-R`)
* add `projects.md` link
* move `R` scripts to `code/` and update links in `debugging-R.md`

ethanwhite:
This all looks good. I'd also recommend getting rid of:

* Debugging Example 'Overkill' Solution
* Functions (Similar Alternate)
* Installing `nose` on Windows
* Scientist's Python Inferno

brymz:
Removed extra files and links.

PR #497, opened 2016-09-13
.gitignore: .jekyll-metadata

brymz:
I am using `jekyll serve --incremental` to optimize rebuilds after local changes. I gain > 20 seconds per change. The process seems to add `.jekyll-metadata`, which we will want to avoid committing.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ Gemfile.lock
 *~
 #*
 *.Rhistory
+.jekyll-metadata


ethanwhite:
Nice. That's a useful thing you just taught me. Thanks!

brymz:
To save a few key strokes you can use `jekyll serve -I`.

brymz:
Also, it can 'break' stuff if the build gets out of order. For example, when you change an `assignments/` you might find you have to run a standard `jekyll serve` to get the markdown formatting of `exercises/` back.

PR #500, opened 2016-09-14
Add external vs local table concept and collect() to `dplyr` exercise and notes

brymz:
As per #499.

### DIFF
--- a/exercises/Dplyr-link-to-databases-R.md
+++ b/exercises/Dplyr-link-to-databases-R.md
@@ -5,34 +5,38 @@ title: Link to Databases
 language: R
 ---
 
-Let's access an SQL database directly from R. Install the `RSQLite` package (and 
-the `dplyr` package if you haven't already).
+Let's access an SQL database directly from R using `dplyr`.
 
 Either use an existing copy of the `portal_mammals.sqlite` database or [download
 a new copy](https://ndownloader.figshare.com/files/2292171). You
-should now be able to link to the surveys table in the database using:
+should be able to link to the `surveys` table in the database using:
 
 ```
 portaldb <- src_sqlite("portal_mammals.sqlite")
 surveys <- tbl(portaldb, "surveys")
 ```
 
-`surveys` will be a `tbl_df`, which means that we won't need to worry about it
-printing out huge numbers of rows in the answers below.
+`surveys` will be a `tbl_sqlite`, which means that the table remains external to 
+the R environment. *Also, we won't need to worry about it printing out huge 
+numbers of rows when we look at it.*
 
-Using this table in the database:
-
-1. Use the `nrow()` function to determine how many records are represented in this 
-   dataset.
-2. Select the year, month, day, and species_id columns in that order
-3. Create a new data frame with the year, species_id, and weight in kilograms 
-   of each individual, with no null weights.
-4. Use the `distinct()` function to print the species_id for each
-   species in the dataset that have been weighed.
-5. Calculate the average size of a *Neotoma Albigula* (`NL`) in this dataset.
-6. Create a new data frame with the number of individuals counted in each year
-   of the study. If you don't know how to count things using `dplyr` I'd
-   recommend checking out the
+1. Use the `nrow()` function to determine how many records are represented in
+   `surveys`.
+2. Select (*and print*) the `year`, `month`, `day`, and `species_id` columns in
+   that order.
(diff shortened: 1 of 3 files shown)


exercises/Dplyr-link-to-databases-R.md
```diff
@@ -5,34 +5,38 @@ title: Link to Databases
...
-   of each individual, with no null weights.
-4. Use the `distinct()` function to print the species_id for each
-   species in the dataset that have been weighed.
-5. Calculate the average size of a *Neotoma Albigula* (`NL`) in this dataset.
-6. Create a new data frame with the number of individuals counted in each year
-   of the study. If you don't know how to count things using `dplyr` I'd
-   recommend checking out the
+1. Use the `nrow()` function to determine how many records are represented in
+   `surveys`.
+2. Select (*and print*) the `year`, `month`, `day`, and `species_id` columns in
```
ethanwhite: I think the "and print" might actually confuse more folks than it will help clarify this for. At this point in the semester they should understand that they need to print out answer.s

exercises/Dplyr-link-to-databases-R.md
```diff
@@ -5,34 +5,38 @@ title: Link to Databases
...
+   that order.
+3. Create a new data frame with the `year`, `species_id`, and weight in
+   kilograms of each individual, with no null weights.
+4. Use the `distinct()` function to print the `species_id` for each
+   species in the dataset that have been weighed. Name this table `species_ids`.
```
ethanwhite: "has been weighed"?

exercises/Dplyr-link-to-databases-R.md
```diff
@@ -5,34 +5,38 @@ title: Link to Databases
...
+5. Determine how many rows are in `species_ids`. *You will have to take this
+   `<derived table>` and make it a local data frame (`tbl_df`) using
+   `collect()`.*
+6. Calculate the average size of a *Neotoma Albigula* (`NL`) in `surveys` stored
+   in a local data frame.
```
ethanwhite: This seems like asking them to do something we wouldn't normally encourage just to practice collect, since in general it will simpler and more efficient to do the work in the database. Or maybe the wording just isn't clear. Are you asking them to get `surveys` into a local data frame before calculating the average (which is how I originally read it) or are you asking them to return the final data from using `collect`?

exercises/Dplyr-link-to-databases-R.md
```diff
@@ -5,34 +5,38 @@ title: Link to Databases
...
+7. Create a local data frame with the number of individuals counted in each year
+   of the study. *If you don't know how to count things using `dplyr`, I'd
+   recommend checking back at the
```
ethanwhite: I've generally started trying to avoid the first person in the materials. Maybe "Take another look at the..."

materials/dplyr.md
```diff
@@ -110,3 +110,23 @@ query <- "SELECT year, month, day, genus, species
           ON surveys.species_id = species.species_id"
 tbl(portaldb, sql(query))
 ```
+
+* Queries and data manipulation functions return similar results with various headings.
+
```
ethanwhite: Can you walk me through you thinking on what you're hoping the students will get out of this new section?

exercises/Dplyr-link-to-databases-R.md
brymz: I am asking to return the final average size value in a local data table.

materials/dplyr.md
brymz: Students should understand the various ways that tables can be derived and stored (external vs. local) using the `dplyr` database interface. One construct using `tbl( sql(query))` is SQL flavored and the other is `dplyr` verbs that manipulate a database table. They should be able to identify the different table output from the external database or a local data frame.

exercises/Dplyr-link-to-databases-R.md
ethanwhite: OK, then I think we should just tweak the language. How about "Calculate the average size of a *Neotoma Albigula* (`NL`) in `surveys` and storing the result in a local data frame."

ethanwhite:
Not sure if you get notified of my in review comments, so pinging here. Let me know if you did/didn't get notified by the other comment.

brymz:
Got the comment and the ping. Already made a similar update to the language.

Issue #502, opened 2016-09-20
Split the dplyr assignment into more exercises


ethanwhite:
Right now the first exercise doesn't really do anything so it's basically 3 exercises but with a bunch of parts each. This is more difficult to keep track of for the students and more of a pain to grade.

ethanwhite:
@brymz - if you happen to have a bunch of time this morning you're welcome to try giving this a go now. I'd need something to review by 1 pm to have a chance of including the change this year. Also totally fine to just put it in the branch for later whenever you get the chance.

brymz:
I can definitely spend my morning on this. Any specific break down that you had in mind?

ethanwhite:
Shrub Volume 3 into 3 problems: 1-6 & 9 (basics), 7-8 (aggregation), 10 (join)
Link to Databases into X problems: 1-4 (basics), 5-7 (more advanced) for this one consider removing the current 6 and adding the Challenge problem I added as the new last sub-problem, but only if you have time to deal with updating the solutions properly as well

brymz:
We're on the same page.

Do you see them being broken up in a way that keeps them grouped (remains sequential, only intro material at the first exercise), or do you envision intro material for each new problem to push the content towards modularity?

ethanwhite:
The more modular the better.

brymz:
Ok. I'm going to prioritize the split. The clean and modularity might have to go `aftersemester`.

ethanwhite:
:+1:

PR #504, opened 2016-09-20
Split dplyr exercises

brymz:
As per #502


brymz:
I think this is what I can get to before 1p EST. I'll be here to follow-up on any comments or changes.

Issue #506, opened 2016-09-20
Replace numbered exercise titles with names (e.g., Shrub Volume 1) 


brymz:
Consider making Shrub Volumes assignment.

Issue #145, opened 2015-07-16
Update Python material to Python 3


ethanwhite:
All necessary modules are now available in Python 3 and it will make teaching easier by removing common points of confusion like integer division. The most common thing we'll been to fix is changing print statements from:

```
print x
```

to

```
print(x)
```

brymz:
Do you want me to back track to the beginning to make this shift while also
formatting? or plan to go through in a pass dedicated to updating to Python
3?

ethanwhite:
Whatever you think is best if fine with me.

brymz:
This ('Update Python material to Python 3') seems like the next long-winded task to get to in between active projects and priority issues. I plan to start in `materials/` and then do `exercises/`. Do you have any feedback or comments before I start? Would you prefer me to work on something else?

brymz:
I think this `2to3` will be very useful (https://docs.python.org/3/library/2to3.html). I am not sure if it will also update `IPYNB`.

ethanwhite:
Should we set up a meeting for next week to strategize on this and other things going forward?

PR #510, opened 2016-09-21
Remove numbers from R exercise titles

brymz:
As per #506.

We may consider this update for 'aftersemester'.


brymz:
This task is complete for all R exercises and ready for review. 

Python and Access exercises can get updated along with #145.

ethanwhite:
I really like this. What's the argument for waiting until the end of the semester?

brymz:
> What's the argument for waiting until the end of the semester?

It was more of a caution that the name and link was modified for a number of the exercises in the previous assignments (`r-intro` & `r-data`). I wasn't sure if that was going to affect the course.

ethanwhite:
As long as the links on the site all still end up in the same place that should be fine. I don't think most folks are writing down the raw links. If that makes sense to you I'll go ahead and merge.

brymz:
And they are referencing exercises by `Problem #` not `Title`. If that's right, good to merge.

ethanwhite:
When turning in assignments you mean?

brymz:
Yes.

Issue #507, opened 2016-09-20
Add note to `dplyr` lesson about Windows installation issues


ethanwhite:
We had about 10% of students today run into an installation issue on Windows. The installation of `dplyr` and `Rcpp` (a `dplyr` dependency) were being blocked by McAfee anti-virus. The solution is to temporarily turn off the streaming anti-virus projection during the installation.

PR #512, opened 2016-09-21
Post 2016 lecture dplyr cleanup

ethanwhite:
Updating with a few lessons learned from teaching the `dplyr` material in Fall of 2016.

### DIFF
--- a/materials/dplyr.md
+++ b/materials/dplyr.md
@@ -10,7 +10,9 @@ language: R
 > * download [`surveys.csv`](https://ndownloader.figshare.com/files/2292172).
 > * download [`species.csv`](https://ndownloader.figshare.com/files/3299483).
 > * download [`portal_mammals.sqlite`](https://ndownloader.figshare.com/files/2292171).
-
+>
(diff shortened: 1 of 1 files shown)


materials/dplyr.md
```diff
@@ -10,26 +10,25 @@ language: R
...
 
 ### Packages
 
 * Main way that reusable code is shared in R
 * Combination of code, data, and documentation
+* R has a rich ecosystem of packages for data manipulation & analysis
 * Download and install packages with the R console:
     * `install.packages("dplyr")`
-* Using a package:
-    * Load all of the functions in the package: `library("dplyr")`
```
brymz: I would think that it would be important to introduce students to this material, at least `library(dplyr)`. Does it belong somewhere else?

ethanwhite: Good catch. Just sloppy, too fast, editing. Fixed.

ethanwhite:
Added fix for #507. Fixed accidental over trimming from earlier. Thanks for the great review.

PR #515, opened 2016-09-23
Add example of speed up from using data in database

ethanwhite:


brymz:
This PR helped clarify the point you make about the 'speed up' using the external database for me and perhaps warrants a bit more discussion. Although, the loading times are clearly in favor of external connections, it seems that calculations take longer externally (as per [this ZevRoss example: section 2](http://zevross.com/blog/2014/03/26/four-reasons-why-you-should-check-out-the-r-package-dplyr-3/)).

ethanwhite:
I don't think we have room for more detailed discussion (in fact I pulled a more advanced example doing aggregation in BBS) so it's either at this level or not at all. Individual benchmarks like the one in the link are generally pretty uninformative, but if you know of something more thorough that shows that dplyr, including the data import, really is faster than databases at this point, then I'm happy to remove the statement entirely. The data import can be a very time cost here, so it definitely needs to be included.

brymz:
I trust your judgement here. I just wanted to let you know of something that I saw in possible conflict.

PR #519, opened 2016-09-27
Add description of how to import tab delimited data to Adult vs. Newborn size exercise

ethanwhite:
Most students won't have seen this yet so best to introduce it in the exercise.


brymz:
Do we remove this language from 'Sexual Dimorphism Exploration' then?

ethanwhite:
I think for modularity probably better to keep it both places.

PR #529, opened 2016-10-04
Switch order of loops and conditions in Programming Fundamentals material

ethanwhite:
Conditionals are easier to teach than iteration, especially in R, so this moves
pairs the conditionals material with functions in the first week of programming
and leaves Loops on their own (with extra bits on problem decomposition and
Bioconductor) for the second week.


ethanwhite:
I know this is a pretty major change, but given that @brymz is interviewing today and I'm teaching this in 4 hours, I'm going to go ahead and merge it. What could go wrong? AmIright?

Issue #464, opened 2016-08-24
Add `dplyr::collect()` to `lectures/r-sql.md`


brymz:
As per #462.

brymz:
Currently, we are teaching `r-sql` integration through `RSQLite`, which uses `dbGetQuery()` in a similar fashion as a `dplyr` pipe ending in `collect()`. In this way, `RSQLite` and `dplyr` overlap and I do not recommend teaching both. We could, however, consider revising the `r-sql` assignment to introduce the database integration components of `dplyr` instead of `RSQLite`. I suppose I would objectively be in favor of a shift to `dplyr` as there are some of the exercises where we use both `RSQLite` (*get's the data table from the database*) and `dplyr` (*further manipulates the local data table*), though perhaps there are important benefits to also teaching `RSQLite`?

We will definitely want to introducing `collect()` in `Dplyr-link-to-databases-R` and update the code and output solutions to reflect that change. I think you are teaching that next week, so I am on it now.

ethanwhite:
I agree that we should just move over the `dplyr` for this. We move that lesson towards: how do we take data out of a database, clean it up, and put it back, all using `dplyr` we can chat more about this at our upcoming meeting.

brymz:
Proposed Revision to `dplyr`:

* "take data out of database": Revise ['Connect and Query'](http://www.datacarpentry.org/semester-biology/exercises/R-sql-connect-and-query-R/).
* a bit more detailed manipulation of database: Revise ['Automate Query'](http://www.datacarpentry.org/semester-biology/exercises/R-sql-automate-query-R/).
* "put it back": Revise ['Export to database'](http://www.datacarpentry.org/semester-biology/exercises/R-sql-export-to-database-R/).
* set-up database and analyze unfamiliar data: Revise ['NEON Databases'](http://www.datacarpentry.org/semester-biology/exercises/Neon-neon-database-R/).
* `tidyr` `gather()` and `separate()`: Revise [`Tree Biomass`](http://www.datacarpentry.org/semester-biology/exercises/Tidyr-tree-biomass-R/).
*`tidyr` `spread()` and maybe `separate()`: Will need to find/create dataset with column that meets `spread()` conditions (e.g., names anther column with values).

PR #532, opened 2016-10-11
R-sql to Dplyr-databases

brymz:
As per #464.


brymz:
This lesson PR is ready for review. 

I revised the exercises from R-SQL to use `dplyr` and further revised 'Manipulate Query' to include a bit more "data wrangling". My intention was to try and fit in a `tidyr` function, but I'm having trouble finding a "real world" scenario for `spread()`. 'Tree Biomass' covers `gather()` and `separate()`. 

I wonder if by adding a 2nd `tidyr` exercises (for 6 in the assignment) that the assignment might get a bit lengthy. Each of these exercises requires some planning/troubleshooting to get the desired result. If you still want more `tidyr` I recommend a basic exercise to use `spread()`. I would probably choose to construction an example dataset to work with.

PR #545, opened 2016-10-25
update knitr materials

davharris:
I added an Rpres presentation to show a bit more about what knitr/Rmarkdown can do. Also added a bit to the examples (tables, more chunk options, inserting results into sentences).

### DIFF
--- a/materials/knitr-examp.Rmd
+++ b/materials/knitr-examp.Rmd
@@ -1,6 +1,6 @@
 ---
 title: "Portal population dynamics"
-author: "Ethan White and David J. Harris"
+author: "Instructor's name"
 date: "December 1, 2015"
 output: pdf_document
 ---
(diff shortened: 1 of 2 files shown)


materials/knitr-examp.Rmd
```diff
@@ -1,8 +1,8 @@
 ---
 title: "Portal population dynamics"
-author: "Ethan White"
+author: "Ethan White and David J. Harris"
```
ethanwhite: Note: this isn't intended to signify the author of the lesson, but to be used to demo the `author` field when teaching the lesson. It could be replaced with `Instructor's Name` or something like that to help indicate this.

davharris: Will change. Thanks

davharris:
Is there any way we could get this merged by the time class starts in 90 min? I think it would make it easier on the students if they don't need to find my fork.  @ethanwhite @brymz

PR #549, opened 2016-11-01
Updates to spatial material

ethanwhite:
1. Split assignment
2. Add more treatment of extract
3. Remove output from code chunks
4. Other minor improvements.


ethanwhite:
@brymz - if you're working today a review in the next hour or so would be awesome. I need the assignment change merged by class, so if you see anything that needs fixing the perfect situation for me would be for you to merge this and then PR in a fix. Thanks and sorry for the last minute rush on this.

brymz:
On it.

brymz:
I will need to update the `solutions/`, but everything else looks great!
Nice revision @ethanwhite.

PR #550, opened 2016-11-01
Update r-spatial file names and solutions with minor revision

brymz:
Follow-up to #549


ethanwhite:
Thanks @brymz! You're awesome.

Issue #524, opened 2016-09-30
Consider adding lecture notes to assignment page


brymz:
I was thinking it would be good to have a place on the site where all of the materials for a topic/lesson/assignment are organized in a single place so that they can be referenced in a single link. Right now, readings and exercises are compiled on the assignments page, but the lecture notes are not. The schedule organizes these three, but does not directly reference a specific topic.

ethanwhite:
Sounds good to me.

PR #553, opened 2016-11-01
Add Lecture Notes Content to Assignment Page

brymz:
As per #524.

BUG REPORT: The 'Lecture Notes' section on `semester-biology/assignments/sql-intro/`, `r-fundamentals-2`, `r-spatial`, `r-git`, and `r-sql`  does not process liquid tags, which breaks the links `{{ site.baseurl }}/...`. This is not a problem on the other `assignments/` pages. I'm head-bashing right now why some and not the others, so any thoughts would be helpful.

Copying one of the files that works and putting it in place of one that doesn't work made no difference.


ethanwhite:
No idea I'm afraid.

So this is WIP at the moment?

brymz:
I think I've got it figured out. Capitalization seems to work. Just putting it together, now.

brymz:
Ok. This is ready for a review.
I am looking for feedback on link functionality and page formatting.

ethanwhite:
This all looks great to me. I won't merge since it's still marked WIP, but if you're happy I think this is ready to go.

Issue #546, opened 2016-10-25
knitr lesson technical issues


davharris:
The lesson went well, but we ran into a few technical hiccups:
- [x] Downloading RMarkdown through the knit button didn't "stick:" students had to `install.packages(rmarkdown)`
- [x] Many students don't have Pandoc/TeX installed; HTML worked more smoothly than PDF
- [x] The example code embedded in the .md file on the webpage should be updated to match the `knitr-examp.Rmd` file (ideally kept in sync automatically for the future)
- [x] The lesson doesn't currently make a clear distinction between the .R extension and the .Rmd extension, so some students were looking for a knit button above a .R file
- [x] The .Rpres and knitr-examp.Rmd files should be easier to find from the website

Also, the lesson still needs:
- [x] To deal with the suggestions @brymz made in #545

brymz:
Are you going to get back to this @davharris? or is this on me?

davharris:
If you have time to deal with it this week, that might be best.  Otherwise, I can definitely get to it eventually.

brymz:
> example code embedded in the .md file on the webpage should be updated to match the knitr-examp.Rmd file (ideally kept in sync automatically for the future)

I wonder about using the `remote_file_content` plugin: https://github.com/dimitri-koenig/jekyll-plugins.

brymz:
Also, we may consider updating this lecture for RStudio V1.0 with notebooks (#491).

ethanwhite:
> example code embedded in the .md file on the webpage should be updated to match the knitr-examp.Rmd file (ideally kept in sync automatically for the future)

Is there are reason we need both of these files?

ethanwhite:
> Also, we may consider updating this lecture for RStudio V1.0 with notebooks (#491).

Agreed, but probably a separate issue.

brymz:
> > example code embedded in the .md file on the webpage should be updated to match the 
> > knitr-examp.Rmd file (ideally kept in sync automatically for the future)
>
> Is there are reason we need both of these files?

I think so. The `MD` is the lecture notes page, while the `Rmd` is the example item that the lecture is about. `.Rmd` are static files according to `jekyll` and won't be rendered into a web page. 

I have integrated material from the `Rpres` to the `MD` and removed the redundant `Rpres`.

PR #556, opened 2016-11-02
Revise knitr Lesson from Feedback

brymz:
As per #546.


materials/knitr-examp.Rmd
```diff
@@ -19,8 +19,8 @@ library(scales)
 Data is from the [Portal project teaching database](http://figshare.com/articles/Portal_Project_Teaching_Database/1314459) published on Figshare. We need the surveys table for our analysis:
 
 ```{r, cache=TRUE}
-# Setting cache=TRUE lets you re-use the results of this code chunk instead of
-# re-calculating or re-downloading it from scratch.
+# Setting cache=TRUE lets you re-use the results of this code chunk in other code
+# chunks instead of re-calculating or re-downloading it each time.
```
ethanwhite: I'm not sure this change exactly right. Code chunks share state automatically. I'm pretty sure that `cache` is primarily operating between knits, not within a single knit.

brymz:
> the demo in a separate file...just happened because last year I happened to commit the Rmd.

The `Rmd` was committed in Aug with the lecture notes: https://github.com/datacarpentry/semester-biology/commit/7e7ca05b587694ab198165cf611f8cbf0e44cf0b

> I write the Rmd file live in front of the class... [from] code stored in the main `MD` document.

Thanks for the clarification. I definitely support live-coding. I will go ahead and move the `Rmd` code chucks to the `MD` and remove the `Rmd`. Are there any specific 'talking points' or stages of development that you would like me to include in the revision?

ethanwhite:
> Are there any specific 'talking points' or stages of development that you would like me to include in the revision?

There is stuff that should be added, but I don't have time to think through it at the moment. Feel free to add something yourself or I'll take care of it later (you can open an issue and assign it to me).

brymz:
As per our discussion, I have integrated the `Rmd` code into the lecture notes `MD` and added talking points. I have included a mention of the R Notebooks to address #557. The lesson can be easily shifted to a Markdown or Notebook focus from there.

I am not happy with the way the code blocks are rendered when using the block spacing method as was used in previous versions of this file. The code blocks are misaligned and there is extra spacing that could get in the way if the live-coder decides to copy/paste the code blocks. I am going to use the background html tags `<pre><code>` to set the blocks 'manually'. I committed the block spacing option in case we decided we preferred that version.

brymz:
Ready to review/merge.

ethanwhite:
This all looks great. Did we end up dropping Dave's `Rpres` material entirely? If so it might be worth adding some of it back to the end demonstrating the ability to make slides using knitr.

brymz:
I made sure the content was merged into the lecture notes. I will add the directions to make an `Rpres`.

brymz:
`Rpres` section added.

ethanwhite:
Thanks @brymz!

Issue #564, opened 2016-11-07
Update Integrating R and SQL lecture materials to match dplyr focus


ethanwhite:
In #532 we moved all of the other material over to `dplyr`, but it looks like we didn't get the lecture material.

brymz:
Working on this now.

PR #566, opened 2016-11-08
Revise r-databases lecture notes

brymz:
As per #564.

### DIFF
--- a/materials/r-databases.md
+++ b/materials/r-databases.md
@@ -1,0 +1,72 @@
+---
+layout: page
+element: notes
+title: dplyr with Databases
+language: R
+---
+
+> Remember to
+>    
+> * download [`portal_mammals.sqlite`](https://ndownloader.figshare.com/files/2292171).
+> * open `portal_mammals.sqlite` in SQLite Manager. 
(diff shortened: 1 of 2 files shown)


materials/r-databases.md
```diff
@@ -0,0 +1,71 @@
...
+
+> Remember to
+>    
+> * download [`portal_mammals.sqlite`](https://ndownloader.figshare.com/files/2292171).
+> * open `portal_mammals.sqlite` in SQLite Manager. 
+> * make sure the copy you are going to use in class does not have the `SpeciesCounts` table or view.
+
+* We've already seen briefly how to work with databases using `dplyr`.
+* Let's get more familiar with the `dplyr` context.
+* There are other database management systems libraries (e.g., `DBI`, `RSQLite`)
```
ethanwhite: `dplyr` is just using these two libraries behind the scenes so the phrasing here is a little confusing. I might say something like "`dplyr` wraps the low level database management libraries to make it easy to interact with databases"

brymz:
> `dplyr` is just using these two libraries behind the scenes.

That would make a whole lot of sense. Language updated to be more clear.

Issue #518, opened 2016-09-27
Replace time-series examples in ggplot lecture


ethanwhite:
The current ggplot lecture analyzes time-series data from Portal. This has a couple of disadvantages:
1. It requires some `dplyr` work up front. I'm not actually sure is this is good (review) or bad (extra time taken up before getting to graphing)
2. Portal is getting tired at this point, time to move on to something new
3. It's a fairly non-standard example of the kinds of bi-variate plots most folks will be used to making

brymz:
How do you feel about the [Gapminder data](https://www.gapminder.org/data/)?

ethanwhite:
I'd rather have something biological if possible. There might be something in there we could use on the disease side of things.

brymz:
Could also continue with the BBS. Other ideas?

ethanwhite:
BBS is tough to quickly get them up and running on. I was thinking about something that more naturally had two or more columns to plot against one another as well as different categories for demonstrating bar plots and histograms.

brymz:
I found that the Andrews LTER Succession Plots Diameter Data fit nicely into the `ggplot` lecture. It also required some `dplyr` to `summarize()` the time series data, but I think that is a realistic process for datasets that biologists will encounter. The code example for the `ggplot` lecture using the Andrews data is https://gist.github.com/brymz/5dc254a6f3b998f9ea057d5a6b74b1ac

brymz:
The idea is to replace the time-series analysis with something directly bi-variate using the Andrews Data.

brymz:
Moving to some data that is directly bi-variate has me considering an introductory figure with `geom_point()` instead of `geom_line()`. Would you still want to introduce `geom_line()` somehow? Even if it means requiring data massaging?

ethanwhite:
I don't think it's crucial, but I'm fine with including it if you have a decent example.

PR #573, opened 2016-11-09
ggplot.md: revise lecture notes to acacia data

brymz:
As per #518.

### DIFF
--- a/materials/ggplot.md
+++ b/materials/ggplot.md
@@ -4,22 +4,17 @@ element: notes
 title: Graphing using ggplot
 language: R
 ---
-
+ 
 > Set up R console:
 
 ```
 library(dplyr)
 library(ggplot2)
 
-surveys <- read.csv("surveys.csv")
-
-ts_data <- surveys %>%
-             group_by(species_id, year) %>%
-             summarize(count = n(), biomass = sum(weight))
-
-dm_ts_data <- filter(ts_data, species_id == "DM")
-do_ts_data <- filter(ts_data, species_id == "DO")
+acacia <- read.csv("http://www.esapubs.org/archive/ecol/E095/064/ACACIA_DREPANOLOBIUM_SURVEY.txt", sep="\t")
 ```
+
+> Get familiarized with [metadata](http://www.esapubs.org/archive/ecol/E095/064/metadata.php) - Acacia drepanolobium Surveys
 				 
 ### Basics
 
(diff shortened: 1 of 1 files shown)


materials/ggplot.md
```diff
@@ -65,110 +68,114 @@ ggplot(dm_ts_data, aes(x = year, y = count)) +
...
-ggplot(surveys, aes(x = weight)) +
+ggplot(acacia, aes(x = CIRC)) +
   geom_histogram()
 ```
 
-* Uses `bin` transform: split weights into bins and count rows in each bin
-* These can be combined with all of other things we've learned
+* Defines `bins` argument
+    * splits circumferences into bins and counts rows in each bin
+* These can be combined with all of other layers we've learned
```
ethanwhite: A number of these commands aren't layers. I can understand not liking "things". How about "approaches" or "commands"?

materials/ggplot.md
```diff
@@ -65,110 +68,114 @@ ggplot(dm_ts_data, aes(x = year, y = count)) +
...
 
 ```
-ggplot(surveys, aes(x = weight)) +
+ggplot(acacia, aes(x = CIRC)) +
   geom_histogram()
 ```
 
-* Uses `bin` transform: split weights into bins and count rows in each bin
-* These can be combined with all of other things we've learned
+* Defines `bins` argument
```
ethanwhite: I'm not sure what "Defines" means here. "Uses `bin` transform" is precise, but if you think it's confusing I'm open to alternative phrasing, but nothing is being defined here.

materials/ggplot.md
```diff
@@ -65,110 +68,114 @@ ggplot(dm_ts_data, aes(x = year, y = count)) +
...
+    * Will default to `ggplot()` mappings
 
 ```
-ggplot(dm_ts_data, aes(x = year, y = count)) +
-  geom_line() +
-  geom_line(data = do_ts_data, aes(x = year, y = count), 
-	color = "red")
+ggplot(acacia, aes(x = CIRC, y = AXIS1)) +
+  geom_point() +
+  geom_point(aes(y = AXIS2), color = "red") +
```
ethanwhite: Let's show a full set of layer arguments here just to help them generalized:

```
geom_point(data = acacia, aes(x = year, y = AXIS2), color = "red")
```

and then maybe add a modification step like:

Since some of the arguments are the same as in the main `ggplot()` call we can use those instead of specifying them

```
geom_point(aes(y = AXIS2), color = "red")`
```

brymz: Done.

materials/ggplot.md
brymz: Went back to "Uses" and added a comment about the default `bins=30` which was the piece I was inferring with "Defines".

materials/ggplot.md
brymz: I went with "`ggplot2` features".

brymz:
Thanks for the feedback! I made revisions as per your comments. Could be ready to go.

materials/ggplot.md
```diff
@@ -65,110 +68,124 @@ ggplot(dm_ts_data, aes(x = year, y = count)) +
...
 
 ```
-ggplot(surveys, aes(x = weight)) +
+ggplot(acacia, aes(x = CIRC)) +
   geom_histogram()
 ```
 
-* Uses `bin` transform: split weights into bins and count rows in each bin
-* These can be combined with all of other things we've learned
+* Uses `bins` argument
```
ethanwhite: There are two different things going on here that are getting mixed together. First, `geom_histogram()` is using the `bin` statistical transformation. Second, there is a `bins` argument that allows you to set the number of bins used by the `bin` transformation. So, I think this first line should be "Uses `bin` transformation" (or transform) and then the bullet point below can be used to mention the `bins` argument.

brymz: Thanks for the clarification. I revised the whole 'Statistical transformations' section based on your feedback to specifically distinguish between `stat_` transformations and shared arguments of the functions.

brymz:
Ready for another look.

ethanwhite:
Thanks @brymz. Looks great!

Issue #541, opened 2016-10-17
Replace, remove, or improve Strings-function-with-choices


ethanwhite:
This question is a perennial source of confusion and it's a little weird to pass back a string like this anyway. We should be able to find something better for this kind of exercise.

brymz:
I'm thinking of introducing a `lm()` or another stats function and printing out a report style `"r-squared = XXX"` or something similar. The choice could be a "significance" indicator like `"***"`.

brymz:
The choice should return a value not a string. Return an r-squared `if()` significant.

PR #575, opened 2016-11-10
Replace Strings-functions-with-choices with Making-choices-choices-with-functions

brymz:
As per #541.


exercises/Making-choices-choices-with-functions-R.md
```diff
@@ -0,0 +1,27 @@
...
+in a tab delimited (`"\t"`) format. Each of the individuals surveyed were
+measured for branch circumference (`CIRC`) and canopy width (`AXIS1`) and was
+identified for the associated ant-symbiont species present (`ANT`). 
+
+1. Write a function that evaluates the linear regression (`lm()`) for the
+   `AXIS1 ~ CIRC` relationship and returns the *r<sup>2</sup>* of the model. 
+
+   *For a `test <- lm()`, get the r<sup>2</sup> using `summary(test)$r.squared`.*
+
+2. Modify the function to take a subset of the data for a given `ANT`.
```
ethanwhite: This is unclear (I'm not actually sure what you're intending). Are they supposed to do the subsetting externally, or are they passing in an argument indicating the species and the function takes care of the subsetting?

exercises/Making-choices-choices-with-functions-R.md
```diff
@@ -0,0 +1,27 @@
...
+3. Nest your regression function into another function that determines
```
ethanwhite: My initial reaction to this is that we should keep this simpler by just using a single function, but I'm open to discussion.

brymz: I was thinking two functions because there are clearly two tasks: 1) get the test results, 2) check the results for significance. [Nested Functions](http://www.datacarpentry.org/semester-biology/exercises/Functions-nested-functions-R/) comes earlier in the assignment, so it would be extra practice of that. *I am continuing the discussion, not strongly advocating for my decision.*

solutions/Making-choices-choices-with-functions-R.txt
```diff
@@ -0,0 +1,3 @@
+[1] "CM - 0.866"
```
brymz: I agree that this is a shortcoming of the lesson. The challenge is to decide under what scenario this programming problem might arise. The two options to me are summary of data as data context or a report of data context. Arguably, this solution fits neither or those scenarios. I propose two alternates:

1) A vector with three values: `c(species, rsquared, significance)`

2) A statement that contains a variable: "Acacia with `species` has a branch ~ canopy relationship that is `significance`, `rsquared`".

brymz: I acknowledge that I have failed to devise an option where the choice returns a value that is NOT a string.

ethanwhite: I like number 1. The challenge is that it can't be a vector since it's mixing numbers and strings, so it would have to be a list. I think this is OK if we give them a chunk of code in the exercise that helps them do this. We talk about lists the following week so this would be a preview.

brymz: Any objection to a `data.frame()` instead of the `list()`?

brymz:
```
  species    r2 significance
1      CM 0.866            S
```

brymz: OR

```
$species
[1] "CM"

$r2
[1] 0.866

$significance
[1] "S"
```

brymz:
Revised as per our discussion to simplify the problem. Ready for another review.

ethanwhite:
Really nice @brymz! I love the move to having them modify the function. In it goes!

Issue #552, opened 2016-11-01
Add more very basic for loops exercises to assignment


ethanwhite:
Some students are still struggling with the concept and adding more basic ones to make sure the fundamentals are solid before moving on would be helpful.

brymz:
Are there particular hang-ups to address?

ethanwhite:
@andrewmarx should be able to clarify.

andrewmarx:
I wouldn't say there are any particular hangups; just a need for more practice with basic concepts to build confidence before incorporating for loops with problem solving. For example, "Print the numbers 3 to 10 with a for loop" and "Print the values in vector x using an index in a for loop." I sent a file to the class with a bunch of examples that could just be converted to actual questions. I'll try to do that sometime over the next week.

brymz:
After all of our recent discussion about collaborative development, I wonder if this is a chance to use/find basic exercises elsewhere as a supplemental 'pre-assignment'

Otherwise, I am thinking of adding two problems: 1) a loop over an index `for(i in 1:10)` and 2) a loop over a vector `for (seq in sequences)`. The problems would use a missing pieces or modify the code style for a slow introduction to the code syntax.

ethanwhite:
I don't mind bringing in something external, but it would have to be licensed for reuse. I would like these to be part of the formal assignment given the degree of confusion. Your suggested additions sound fine to me.

brymz:
Didn't find much that I thought could be useful from a quick keyword search. I'll get started on the exercise additions.

PR #594, opened 2016-11-22
Add basic loops exercises

brymz:
As per #552


exercises/Loops-basic-vector-R.md
```diff
@@ -0,0 +1,30 @@
...
+`name`:
+
+```
+library(stringr)
+pet <- "spot"
+pet_letters <- str_split(pet, "")[[1]]
+first <- toupper(pet_letters[1])
+others <- str_c(pet_letters[-1], collapse = "")
+name <- str_c(first, others, sep = "")
+```
```
ethanwhite: A simpler version of this would be:

```
library(stringr)
pet <- "spot"
name <- str_to_title(pet)
```

The problem is that since `stringr` is vectorized then this solves the overall problem:

```
library(stringr)
pets <- c("spot", "gigantor", "fluffy")
names <- str_to_title(pet)
```

which is maybe why you took the more complicated approach.

exercises/Loops-basic-index-R.md
```diff
@@ -0,0 +1,23 @@
...
+The following code prints the first five whole numbers one line at a time:
+
+```
+for (i in 1:4){
+  print(i)
+}
+```
+
+1. Modify the code to print five orders of magnitude (*from one to ten
+   thousand*) one line at a time.
```
ethanwhite: I really like the general idea here, but I didn't understand what you were asking for with these "orders of magnitude" problems until I looked at the answer. The other issue I see here is that lots of bio folks won't easily think of the math required to understand how to do this (I speak from experience having removed problems in the last that assumed basic understanding of logarithms).

Maybe start with something simple "Three times each of the numbers from 1 to 10"? I like number 4. It might also be useful to do something that would get them to add at least one more line inside the loop. One other thing that might be useful to include is having them change the name of the variable so that they understand that `i` isn't special.

exercises/Loops-basic-vector-R.md
ethanwhite: `stringr` in general seems to be a mine field for this sort of thing (having learned this lesson the hard way several times myself. What about just using something like `cat` or `paste` instead? Making lines like, "My pet's name is spot"?

brymz:
Great suggestions, @ethanwhite!  I revised the exercises with your ideas in mind. I also added another task to 'Basic Index' that guides students to create a vector with a loop.

ethanwhite:
👍

PR #596, opened 2016-11-29
schedule.md: rm hurricane wk

brymz:


ethanwhite:
We'll need to add an additional project week to get back up to a full semester I think.

brymz:
Done.

Issue #565, opened 2016-11-07
Check solution to phenology from space


ethanwhite:
I (and several students) get a subtly different graph.

We solved it this way (just the HARV data in the example):

```
library(raster)
library(dplyr)
ndvi_files = list.files("data/HARV_NDVI/",
                         full.names = TRUE,
                         pattern = "HARV_NDVI.*.tif")
ndvi_rasters <- stack(ndvi_files)
avg_ndvi <- cellStats(ndvi_rasters, mean)
avg_ndvi_df <- data.frame(samp_period = seq_along(avg_ndvi), ndvi = avg_ndvi)
ggplot(avg_ndvi_df, aes(x = samp_period, y = ndvi)) + geom_point()
```

brymz:
It looks like the difference in solutions is that I used the `mean` of the `plot_locations` at each site instead of the `mean` of the whole raster image. The language is unclear in the exercise, so a revision could go either way.

ethanwhite:
Ah, I see. What do you think about modifying it to include both as separate tasks. That way we can introduce `cellStats` but also reinforce extracting data for sites (and doing so repeatedly).

brymz:
That's a great idea. Is that an `aftersemester` update? Do I update the output solution to your code for now?

ethanwhite:
Yes and Yes.

PR #597, opened 2016-11-29
Expand Phenology from Space to incluce two mean calculation tasks

brymz:
As per #565.

Warning: This is an `aftersemester` patch directly to `gh-pages`.


brymz:
Code solution: https://github.com/weecology/datacarp-semester-biology-exercise-solns/pull/135

ethanwhite:
👍

PR #619, opened 2017-01-06
Fixing typos

atyre2:
Is there an easier way to submit these other than creating a whole new branch and redoing them? pull requests straight from my gh-pages branch include *alot* of stuff specific to my course.

### DIFF
--- a/about/audience.md
+++ b/about/audience.md
@@ -40,6 +40,6 @@ like it would be a lot easier if it could be automated.
 
 **Paul:** A faculty member with little computer background who realizes
 that computing has become so central to his area of research that he
-needs to at least have a rough idea of how his students are work the
+needs to at least have a rough idea of how his students are working the
 magic they do for projects with a lot of programming. Even better he'd
 like to get to the point where he can really engage and help out.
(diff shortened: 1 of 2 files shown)


ethanwhite:
Thanks for these @atyre2!

The combination of customized courses and contributing changes upstream is definitely a little tricky. One thing that will work in a reasonable number of cases is to keep the upstream changes in their own commits and then cherry-pick them into a branch for a PR. E.g.,

```
<make changes>
git add stuff.md
git commit -m "Changes for just your class"
<make changes>
git add stuff.md
git commit -m "Upstream changes"
git log
<note hash(es) for upstream changes commit(s)>
git checkout -b upstream-changes
git cherry-pick firsthash secondhash
git push origin upstream-changes
<open PR>
```

If you run into cases where this sort of thing doesn't work, or if the above approach gives you trouble, I'd be happy to  chat more.

ethanwhite:
There was a bug in my original example that I've just cleaned up (so that you don't end up with all of your changes already in the PR branch. Let me know if this makes sense and seems like a helpful approach. If it does I'll add it to the docs on the site.

PR #623, opened 2017-01-30
Fixing variable in *Combining different data sources*

atyre2:
Success! small victories ...


ethanwhite:
Thanks @atyre2! Sorry it was so much work.

PR #626, opened 2017-04-22
Fix link to SQL readings on Readings page

ethanwhite:
These links changed after we updated the structure of the first two weeks
but we missed these links.


ethanwhite:
Merging since these are broken on the live site. Tested locally.

PR #627, opened 2017-04-22
Fix links to external readings

ethanwhite:
A number of Data Carpentry lesson links were apparently moved and not
forwarded over the last fews. This updates them to working versions.


ethanwhite:
Merging since these are actively broken. Already tested locally.

PR #628, opened 2017-05-23
Fixing broken link in the instructions for customizing a page

mikoontz:
Link to the site-structure file should, I think, point to the site-directory file now. It currently returns a 404 error, so if it's not the site-directory, it should be something else.


brymz:
Thanks for the fix, @mikoontz! Hope you found the `docs/` otherwise accessible and informative.

mikoontz:
Yes, quite! Thanks for putting together such a great resource.

PR #635, opened 2017-08-30
Minor updates to SQL lectures

KristinaRiemer:
Hopefully I didn't mess this up * fingers crossed *


KristinaRiemer:
I didn't realize commits need to be on different branches to be part of different pull requests, so the two most recent commits are for adding submission details. Oops. 

I added info about assignment file types to the checklist file, then changed the name of that file to something more general ("Assignment Submission & Checklist"). 

I kind of think this page shouldn't be linked to on the main page, because it's not as important as the other links, and should only be linked to via the top of the schedule page?

PR #639, opened 2017-09-04
Add short lecture on making sure code runs

ethanwhite:
Discuss how to make sure code runs to help reinforce key points about reproducibility early and decrease the number of issues with grading.


ethanwhite:
I think this is probably a valuable addition for Friday if you have time, but see what you think.

KristinaRiemer:
Is there a way to look at the website with these changes, or does that only work locally?

KristinaRiemer:
I'm not going to merge this until after class today, just so the website doesn't change during class and put people off-kilter.

ethanwhite:
:+1:

PR #652, opened 2017-09-20
Change intro R structure

KristinaRiemer:
Sorry the second commit has so much stuff in it, it didn't occur to me to break it up into several commits until I was done.


ethanwhite:
Nice work!

PR #662, opened 2017-09-26
Add second correct answer to dplyr problem 6.3

KristinaRiemer:


ethanwhite:
Thanks! We should add the code to handle this to the solutions repo as well.

PR #663, opened 2017-09-26
Add ggsave lecture material & exercise task

KristinaRiemer:


exercises/Graphing-sexual-dimorphism-data-manipulation-R.md
```diff
@@ -60,3 +60,4 @@ resulted in a lot of plots, but many of them had low sample sizes.
    difference for each family, only including families with more than 25
    species.
 
+4. Save the figure from task 3 as a .jpg file with the name `sexual_dimorphism_histogram`.
```
ethanwhite: We should probably include the file extension in the filename since they'll need to put the full name into `ggsave`.

KristinaRiemer: I thought about doing that, but was hoping they would figure that out themselves. Do you think they would leave the file extension out?

ethanwhite: That's my guess.

PR #664, opened 2017-09-27
Split up week 7 into separate projects and version control weeks

KristinaRiemer:


ethanwhite:
My read through looks good. Did you check a local build to make sure that links to the exercises and answers didn't accidentally get broken in the renaming of things? (I think it should all be good, but it's always worth testing these)

KristinaRiemer:
I looked at some of them, but not all of them in a systematic way. I'll do that tomorrow.

ethanwhite:
That's probably fine. I think they'd likely all break or none of them. Go ahead and merge this when you're happy.

KristinaRiemer:
Because they won't have been introduced to git when they initially turn in their project proposal, I moved creating the project repo into its own assignment at the end of the version control assignments. @ethanwhite, can you make sure that looks okay?

PR #667, opened 2017-09-29
Make visual link formats consistent

ethanwhite:
Links that were composed of package names appeared like code chunks because
of the use of ` for package names. This was confusing to some students who
didn't realize they were links. This removes the ` in links to avoid this
confusion and present a consistent appearance for all links.


KristinaRiemer:
So maybe that's why some people didn't do the reading? It's fine to have it consistent though.

PR #668, opened 2017-09-30
Project structure (week 7) redo

KristinaRiemer:
Are the lecture notes too long/detailed, and did I miss anything important? 

I also don't know how to get rid of the new line that shows up after every list with indented subsections, but it looks terrible.


ethanwhite:
Looks good to me. I think length is about right. I'd probably try to condense some of the wording a bit just to make it easier to grok at a glance, but I typically make those kinds of improvements when revising year-to-year. This is a great first version.

The line spacing is a feature of the theme. That can be changed, but it's a separate issue and you'd need to dig into CSS to figure it out.

I'm going to go ahead and merge this.

PR #669, opened 2017-10-04
Change data and code for git lecture

KristinaRiemer:
As the git week is in the middle of the semester instead of the end, the code (based on writing functions) is no longer appropriate for what these students have learned. I changed it to examples using both dplyr and ggplot to give them more exposure. 

Because I was having to do that anyways, I found an ecological dataset on fish size and growth rate to replace the house elf made-up data. 

If all of this seems okay, I'll change the exercises correspondingly after this PR is merged.


ethanwhite:
Looks great!

PR #701, opened 2017-10-24
Simplify Writing Functions exercise and move to front of assignment

ethanwhite:
This exercise was somewhat redundant with the nested functions exercise and
interrupted the flow between the two dinosaur exercises. This gives us a really
simple first exercise before moving into more complex material. It also allows
the first dinosaur exercise to reinforce function basics after a bit more
material is introduced.

### DIFF
--- a/materials/functions-R.md
+++ b/materials/functions-R.md
@@ -56,7 +56,7 @@ calc_shrub_vol(0.8, 1.6, 2.0)
 shrub_vol <- calc_shrub_vol(0.8, 1.6, 2.0)
 ```
 
-> * Do [Exercise 1 - Use and Modify]({{ site.baseurl }}/exercises/Functions-use-and-modify-R), Tasks 1-2.
+> * Do [Exercise 1 - Writing Functions]({{ site.baseurl }}/exercises/Functions-writing-functions-R).
 
 * Treat functions like a black box.
     * Can't access a variable that was created in a function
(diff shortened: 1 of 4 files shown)


materials/functions-R.md
```diff
@@ -56,7 +56,18 @@ calc_shrub_vol(0.8, 1.6, 2.0)
...
+> * Do [Exercise 1 - Writing Functions]({{ site.baseurl }}/exercises/Functions-writing-functions-R).
+=======
+* Walk through function execution
+    * Call function
+	* Assign 0.8 to length, 1.6 to width, and 2.0 to height inside function
+	* Calculate volume
+	* Send the volume back as output
+	* Store it in `shrub_vol`
+
 > * Do [Exercise 1 - Use and Modify]({{ site.baseurl }}/exercises/Functions-use-and-modify-R), Tasks 1-2.
```
KristinaRiemer: This is Exercise 2 now, right?

materials/functions-R.md
```diff
@@ -56,7 +56,18 @@ calc_shrub_vol(0.8, 1.6, 2.0)
 shrub_vol <- calc_shrub_vol(0.8, 1.6, 2.0)
 ```
 
+<<<<<<< Updated upstream
```
KristinaRiemer: Is this line and lines 61 and 70 supposed to be here? I don't really understand their purpose, this looks sort of like a merge conflict.

ethanwhite:
Thanks goodness for review! I fixed these (I think). And started moving in a direction that I think is more useful of not listing exercise numbers (which are more likely to change) and listing just their names instead.

KristinaRiemer:
I think that's generally a good idea.

PR #702, opened 2017-10-24
Actually have students calculate something in Writing Functions

ethanwhite:
Also update the answer to not include rounding.


KristinaRiemer:
I'm a little annoyed that I didn't notice this.

PR #706, opened 2017-10-27
Add an exercise that just creates an if/else statement

ethanwhite:


ethanwhite:
Merging since I'm using this in 45 minutes.

PR #710, opened 2017-10-30
Improve first third of programming fundamentals 2

KristinaRiemer:
Changes made up to the "Looping in dataframes" section. The lecture material and exercises were not synced up, so I tried to fix all of that. Also transitioned to species names as more biologically relevant examples and removed exercise numbers.

### DIFF
--- a/exercises/Loops-basic-index-R.md
+++ b/exercises/Loops-basic-index-R.md
@@ -13,16 +13,17 @@ for (i in 1:5){
 }
 ```
 
-1. Modify the code to print three times each of the numbers from 1 to 5 one line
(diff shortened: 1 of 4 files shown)


ethanwhite:
These are all improvements over the existing material so going ahead and merging.

The parent exercise for Exercise 3 has changed the context, so the wording at the beginning of that exercise needs to be changed. This could be as simple as deleting the first sentence.

KristinaRiemer:
It's already on my list of things to do, I'm just working through the lecture notes in order and now 3 is beyond what I covered here.

PR #718, opened 2017-11-02
Rest of changes to PF2 lecture

KristinaRiemer:


materials/for-loops-R.md
```diff
@@ -208,140 +201,120 @@ waterbirds_cap_list
...
-date,time,lat,long
-2016-01-01,4:20,26.16,-35.28
+for (i in 1:nrow(waterbirds)){
+  print(paste(waterbirds$sci_name[i], "is a", 
+              waterbirds$common_name[i]))
+}
 ```
 
-* To get a list of all these files in our `data` directory use `list.files()`
+* Less memory to create initial empty dataframe
```
ethanwhite: Not sure what the goal of this line is. Pre-allocating will end up with the same size at the end, which is probably all that matters, and the copying will actually end up hitting double the memory in the last step. In general I'd recommend pre-allocating the size of the dataframe, but I think that's an extra level of complexity so I haven't taught it that way.

materials/for-loops-R.md
```diff
@@ -208,140 +201,120 @@ waterbirds_cap_list
...
+* Creates copy of dataframe when adding rows
```
ethanwhite: This might also be too complicated for day 1 loops. Good responses to have available for questions though.

KristinaRiemer:
@ethanwhite both of those new bullets were just my attempt to condense what was there before, which was the following: 

* We can also use the index to store the output by pre-allocating memory
* This is a lot faster than adding rows because it doesn’t copy the data frame every time through the loop

Maybe I misinterpreted what these two comments mean?

PR #757, opened 2018-08-21
R SQL lecture

KristinaRiemer:
Updated some of the code that doesn't work anymore because of changes in how dplyr does SQL and removed all dependencies on SQLite Manager.


ethanwhite:
Thanks @KristinaRiemer! Looks great. Sorry it took so long for me to look it over.

Issue #755, opened 2018-08-17
Create a data or spreadsheet language


ethanwhite:
Some topics really are just about data related things, not SQL. Either rename to "spreadsheets" or the like or just use "Data"

KristinaRiemer:
What about changing "Language" to "Topic", and including "Data management" along with R, SQL, etc.? Would this just have to be changed in the xx-template.md files?

ethanwhite:
I like that, though we'll have to see what something that long does to the layout of http://datacarpentry.org/semester-biology/schedule/. I think "data" might work if the longer form is problematic. It should be changed in the template files, but also anywhere else that variable gets used. E.g., on the schedule page and lecture materials page, etc.

KristinaRiemer:
I'll take a crack at it.

ethanwhite:
Awesome. Thanks!

KristinaRiemer:
"Topic" is already being used as part of the yaml, so I'm going to use "Focus" instead.

ethanwhite:
It also looks like `Topic` is barely used. From a quick glance it looks like it might just be used an as extra level of title on the individual exercise pages. If that's the case then we could just replace the current usage. The current use of `Topic` is potentially usefuI for selecting exercises to swap in and out, which would argue for keeping it, but I'm a little concerned that `Focus` is a pretty nebulous term. So, no strong opinions here, just throughts.

KristinaRiemer:
Yeah, I also think that the term "Focus" is too vague, but couldn't think of anything better. I can co-op "Topic" if you're okay with that?

KristinaRiemer:
I also still don't really understand how the website is structured. I tried to just replace as many instances of "Language" with "Topic" as I could find in the template and materials/exercises .md files. Is there a particular high-level file that that should be changed in?

ethanwhite:
Yeah, most of sites underlying mechanics occurs in the files in the `_includes` and `_layouts` directories.

ethanwhite:
Happy to chat more about what's going on there after you take a look if that would be useful.

PR #758, opened 2018-08-23
Change categories

KristinaRiemer:
An attempt to address #755. 

The first commit was fine, but the second broke the assignment page and didn't change the Language header. I'm not sure what else to do from here @ethanwhite. 

Just for future reference, I used this perl code in the command line for these, and had to go into each subfolder to apply to all files: `perl -pi -w -e 's/SEARCH_FOR/REPLACE_WITH/g;' *.*`


ethanwhite:
The key issue here is that the current PR changes all of the metadata in the YAML, but doesn't change how it is handled by the code that generates the site.

This code is primarily in the `_includes` and `_layouts` directories. For example, in `_includes/assignments.html` the repeated uses of `*.language` are referring to the `language` metadata item in the YAML. Likewise, `_layouts/exercise.html` includes the use of `page.topic`, which is referring to the `topic` metadata item.

If I was making a change like this I'd typically start by looking through and editing those files, then experimenting with one or two exercise/materials/assignment files to see if things are working as expected, and only then doing the full find and replace.

PR #774, opened 2018-09-17
Add help on intermediate steps to phenology exercise

ethanwhite:
To make this works requires a lot of little R tricks that many students
won't know. Help them with those so they can focus on the spatial work.


ethanwhite:
@KristinaRiemer This one will also require an update to the solution (just the first part). I thought it was better to have them do the exercise in the way that's most inline with how ggplot should be used, hence the change.

KristinaRiemer:
This looks good, I'm going to merge it and do a new PR to update the first solution figure.

PR #780, opened 2018-09-19
Add introduction to projects material to 2018

ethanwhite:
Includes some updates to the material and the addition of material on style in R.

### DIFF
--- a/assignments/project.md
+++ b/assignments/project.md
@@ -1,9 +1,10 @@
 ---
 layout: page
 element: assignment
-title: Projects Structure
+title: Computational Projects
 language: R
-exercises: ['Project Proposal']
+exercises: ['Megafaunal Extinction']
+points: [60]
 ---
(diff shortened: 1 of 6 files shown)


ethanwhite:
I'm opening this for comment. We have material to teach on programming style, problem decomposition, and computational project structure. It seems like some of this should be introduced next week as the students start to tackle bigger projects, including at least some version of problem decomposition, but some of it should also be introduced during "Building Up Complex Programs".

Do folks have thoughts on the right breakdown?

ethanwhite:
Sounds good. So maybe that and some problem decomposition next week and save style for 'Building Up Complex Programs' along with a refresher on problem decomposition and some material on debugging?

ethanwhite:
Apparently I accidentally pushed this to master last week. Sigh. I'll update based on this discussion.

PR #781, opened 2018-09-21
Add new climate space exercise

KristinaRiemer:


KristinaRiemer:
@ethanwhite I agree with all of your suggestions, see 874a7b5 for those.

ethanwhite:
Looks great! Thanks!

PR #783, opened 2018-09-25
Split climate space into general question and tips

ethanwhite:
This helps it read a bit more like a bigger problem.

Also moves combining all data together on a single plot/set
of plots to a challenge option at the end.


KristinaRiemer:
I like the bulleted points, the two parts were getting very lengthy.

KristinaRiemer:
The solutions figures still need to be updated to show the three plots for each species @ethanwhite?

PR #787, opened 2018-10-01
Create new followup climate space exercise, with solutions

KristinaRiemer:
Okay, here's the first version of the new exercise. I'll also be submitting a PR for the solution in a second. 

Should it go between the current third (Nested Functions) and fourth (Choice Operators) exercises for this week?


ethanwhite:
Looks great. Should the order of 1 & 2 be reversed since they need to first get the data and then plot it?

ethanwhite:
I'll add it into the assignment once it's merged since I'm changing some other things in the assignment right now.

KristinaRiemer:
I figured having them do the plotting function first because it's easier? And it's realistic to refactor different parts of the code at different times, not necessarily in order. I'm happy to switch them around though.

ethanwhite:
I totally get where you're coming from, but I suspect that the out of order refactoring will confuse novices so let's go with the natural data -> plot order.

ethanwhite:
Awesome! Thanks. In it goes.

PR #802, opened 2018-11-06
Expand knitr to full assignment

ethanwhite:
Teaches the basics of knitr plus reinforces many other aspects of the course.


ethanwhite:
Merging for class.

PR #847, opened 2019-09-20
Update Graphing-acacia-ants-data-manip-R.md

andrewmarx:
Part 4 depends on the filtering in part 3. Someone was having issues with 4 because this wasn't clear enough; they were piping their filtered data into ggplot in part 3, so they never updated it.


ethanwhite:
Thanks @andrewmarx!

PR #856, opened 2019-10-28
Replace ** with ^ for exponentiation

andrewmarx:
`**` is a holdover from S.

`?'**'` `** is translated in the parser to ^`


ethanwhite:
Thanks!

PR #862, opened 2019-11-12
Update instructions and add new checks for git-RStudio installs

ethanwhite:
* Provide instructions for fixing path issues on Windows
* Extend the install check to ensure that authorization is working


ethanwhite:
@andrewmarx - can you take a look at this and see if it covers all of our issues from today. Happy to add more, but this is what I could remember once I finally sat down.

andrewmarx:
It looks good. It doesn't cover every little thing like permissions and credentials, but that would over-complicate the page for minimal gain

PR #878, opened 2020-08-12
Force use of netlify default Ruby version

ethanwhite:


ethanwhite:
Replaced with commit in #879

PR #884, opened 2020-08-31
Numbered list

garezana:
Edited for clarity and change 2.2046 to 2.20462 to remove approximately.

Please delete this line and the text below before submitting your contribution.

---

Thanks for contributing! If this contribution is for instructor training, please send an email to checkout@carpentries.org with a link to this contribution so we can record your progress. You’ve completed your contribution step for instructor checkout just by submitting this contribution.

Please keep in mind that lesson maintainers are volunteers and it may be some time before they can respond to your contribution. Although not all contributions can be incorporated into the lesson materials, we appreciate your time and effort to improve the curriculum.  If you have any questions about the lesson maintenance process or would like to volunteer your time as a contribution reviewer, please contact The Carpentries Team at team@carpentries.org.

---

### DIFF
--- a/exercises/Expressions-and-variables-basic-variables-R.md
+++ b/exercises/Expressions-and-variables-basic-variables-R.md
@@ -14,10 +14,8 @@ mass_g <- mass_kg * 1000
 mass_g
 ```
 
-Modify this code to create a variable that stores a body mass in pounds and
-assign it a value of 3.5 (about the right size for a
-[Desert Cottontail Rabbit – *Sylvilagus audubonii*](https://en.wikipedia.org/wiki/Desert_Cottontail)).
-Convert this value to kilograms. There are approximately 2.2046 lbs in a
-kilogram, so divide the variable storing the weight in pounds by 2.2046 and
-store this value in a new variable for storing mass in kilograms. Print the
-value of the new variable to the screen.
+1. Create a variable to store a body mass in pounds. Assign this variable a value of 3.5 (about the right size for a [Desert Cottontail Rabbit – *Sylvilagus audubonii*](https://en.wikipedia.org/wiki/Desert_Cottontail)).
+2. Convert the variable from body mass in pounds to body mass in kilograms and assign it to a new variable.
+Tip: You can do this by dividing the variable storing the weight in pounds by 2.20462. 
+3. Print the value of the new variable to the screen.
+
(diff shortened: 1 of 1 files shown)


ethanwhite:
Thanks @garezana. We use numbered lists to indicate subexercises (e.g., https://datacarpentry.org/semester-biology/exercises/Expressions-and-variables-basic-expressions-R/), so we need to avoid that structure for this exercise.

I think we also need to reference the original code in some way, because otherwise students will wonder why they were shown that code in the first place.

What do you think about something like:

> Create a similar piece of code to convert a mass in pounds to kg.
> * The first line should
> * The second line should
> * The third line should

I also think we should also remote the parenthetical statement about the rabbit (I used to think this kind of detail helped group things, but I think they actually get in the way for most beginners).

I'd probably also shorten the second bullet to something like:

> Convert the variable from body mass in pounds to body mass in kilograms (by dividing it by 2.20462) and assign it to a new variable.

Lastly, if we change the divisor, this will change the solution, so we need to update the code solution in the solutions repo and the text solution in the solutions directory in this repo (I think it would also be fine to just leave the extra digit to avoid the extra work, but I'm happy either way).

garezana:
>  Create a similar piece of code to convert a mass in pounds to kg.
>
> -  The first line should
>  -  The second line should
>  -  The third line should

Adding the reference does make it clearer. 

> I also think we should also remove the parenthetical statement about the rabbit (I used to think this kind of detail helped group things, but I think they actually get in the way for most beginners).

I think it makes it seem more relevant for the field and gives the exercise some context. Maybe remove the size word  as it makes me think of volume and not mass.

 Maybe something like:
> Create a variable to store a body mass in pounds. Assign this variable a value of 3.5 (an appropriate mass for a [Sylvilagus audubonii](https://en.wikipedia.org/wiki/Desert_cottontail)).

> Create a variable to store a body mass in pounds. Assign this variable a value of 3.5

Removing it makes it seem too x=value.


> Lastly, if we change the divisor, this will change the solution, so we need to update the code solution in the solutions repo and the text solution in the solutions directory in this repo

Originally the text said: 
> Convert this value to kilograms. There are approximately 2.2046 lbs in a kilogram, so divide the variable storing the weight in pounds by 2.2046 and store this value in a new variable for storing mass in kilograms. 

Tried to make the instruction shorter by removing the: "store value in a variable for storing" and changing "approximately  2.2046" to 2.20462 . Which was what submitted yesterday with a consistency note and why i asked on dependencies yesterday to not do this mistake again.

garezana:
Pushed again integrating suggestions.

ethanwhite:
Looks great! Thanks!

PR #886, opened 2020-09-13
Fix exercise description of how to calculate area

ethanwhite:


ethanwhite:
@andrewmarx notifying you that a student caught this error in the question. I'm merging this in since most students won't have started on this yet, but if anyone does width * height on the assignment they should get full credit.

Issue #915, opened 2021-11-03
Implement broken link checking


ethanwhite:
A number of broken links were reported in https://github.com/openjournals/jose-reviews/issues/139. We should implement link checking using GH Actions and fix the broken links identified.

andrewmarx:
There's a 'html-proofer' Ruby gem that might work for this. Here is a short tutorial explaining how to implement it as a github action: https://clementbm.github.io/github%20action/jekyll/link%20checking/2020/05/31/automatically-validate-links-on-jekyll-website.html

I'm not familiar enough with Jekyll/Ruby/Github actions to figure out what needs to be adjusted from that tutorial for the course website, but I could maybe do it in the next month or two when I have time if no one else decides to tackle it before then.

ethanwhite:
@andrewmarx - definitely feel encouraged to give this a go if you have the time/energy. I thought I had a quick and easy solution but it failed. I used `linkchecker` to do a manual check and have fixed all of the issues, but would love to have this running on builds and a weekly cron job to prevent the build up of broken links.

ethanwhite:
Removing `jose-publication` milestone since the broken links themselves have been fixed satisfying the requirements for https://github.com/openjournals/jose-reviews/issues/139

ethanwhite:
Completed by @andrewmarx in #935

PR #920, opened 2021-11-04
Add link checking using lychee

ethanwhite:


ethanwhite:
Closing since this isn't working as expected. The approach identified in #915 by @andrewmarx is a better approach.

PR #931, opened 2021-12-17
Test new link checker - no broken links

ethanwhite:


ethanwhite:
@andrewmarx - looks like we're not working yet. looks like an installation issues before we get to the actual checks.

PR #933, opened 2021-12-17
Update checklinks.yml

andrewmarx:


andrewmarx:
Closer, but not quite there. Internal relative paths aren't being identified correctly

PR #935, opened 2021-12-17
Link checker fix

andrewmarx:


andrewmarx:
Almost there I think.

The first "bad link" on the 404 page is a false positive due to a missing trailing / in the link. I saw somewhere discussions about this in html-proofer's issues. I think there is a workaround for it, I just need to find it again

After that, the R conditional assignment has 3 true positives (the 3 lecture links).

Haven't hand checked the rest.

andrewmarx:
It's basically working now. Before, broken links would trigger an error stopping the workflow. Now it will continue to success even with broken links. The broken links can be viewed by looking through the details in the check, but maybe it would be better to output the results somewhere more visible.

Some notes:
- It doesn't check just links, it also checks scripts to make sure internal references are working and that external scripts are loading. That's why there are some javascript related issues being reported.
- There's a bunch of options in html-proofer that might be worth exploring.

ethanwhite:
Thanks @andrewmarx!

PR #1024, opened 2023-11-27
fix join example

BeastyBlacksmith:


ethanwhite:
Thanks for the fix @BeastyBlacksmith!

